PR #12, opened 2022-12-09
Update README

yim-lee:
Document async/await API


tomerd (approved):

README.md
```diff
@@ -1,128 +1,232 @@
...
+}
 ```
 
 ```swift
-  session.query("select * from table ...").map { (rows: Rows) in
-    // rows is a sequence that one needs to iterate on
-    rows.map { row in
-      ...
-    }
+session.query("select * from table ...").map { (rows: Rows) in
```
tomerd, line 205: parenthesis not required

PR #23, opened 2023-04-04
Swift 5.8 docker images are available

yim-lee:


tomerd:
should we add 5.9 nightly if not already there?

tomerd (approved):

PR #29, opened 2024-02-22
Symlink header files

yim-lee:
Motivation:
On some platforms `swift build` is failing with error:

```
 public headers ("include") directory path for 'CDataStaxDriver' is invalid or not contained in the target
```

Modifications:
Copying https://github.com/swift-server/swift-kafka-client's structure, adding `CDataStaxDriver/include` and symlinks to header files under `datastax-cpp-driver/include`.

rdar://123389483

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -137,7 +137,6 @@ let package = Package(
                 "./datastax-cpp-driver/src",
                 "./custom/src",
             ],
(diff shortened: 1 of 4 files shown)


yim-lee:
@swift-server-bot test this please

yim-lee:
Turns out the said problem is due to missing git submodules. I won't merge this PR then.

PR #30, opened 2024-06-04
Make Statement.setPagingSize method public

virtan:
Make Statement.setPagingSize method public

### Motivation:

In some cases we may need to change paging size to reduce chances of read timeouts

### Modifications:

The method setPagingSize() existed but wasn't public so was inaccessible for clients

### Result:

With this change it would become public and let a client to set custom paging sizes on per statement basis

### DIFF
--- a/Sources/CassandraClient/Statement.swift
+++ b/Sources/CassandraClient/Statement.swift
@@ -146,7 +146,8 @@ extension CassandraClient {
             return cass_statement_bind_collection(self.rawPointer, index, collection)
         }
 
-        func setPagingSize(_ pagingSize: Int32) throws {
+        /// Sets the paging size of the returned paginated results.
+        public func setPagingSize(_ pagingSize: Int32) throws {
             try checkResult { cass_statement_set_paging_size(self.rawPointer, pagingSize) }
         }
 
--- a/Sources/CassandraClient/Statement.swift
+++ b/Sources/CassandraClient/Statement.swift
@@ -147,8 +147,8 @@ extension CassandraClient {
         }
 
         /// Sets the paging size of the returned paginated results.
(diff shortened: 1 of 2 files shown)


Sources/CassandraClient/Statement.swift
```diff
@@ -131,7 +131,8 @@ extension CassandraClient {
             return cass_statement_bind_collection(self.rawPointer, index, collection)
         }
 
-        func setPagingSize(_ pagingSize: Int32) throws {
+        /// Sets the paging size of the returned paginated results.
+        public func setPagingSize(_ pagingSize: Int32) throws {
```
glbrntt, line 135: Can we use `Int` instead here? It should be preferred for public APIs.

virtan:
There is a variant of `execute()` method which also expects `pageSize` in `Int32`. It is public already, so changing it may introduce some backward compatibility issues.

glbrntt:
> There is a variant of `execute()` method which also expects `pageSize` in `Int32`. It is public already, so changing it may introduce some backward compatibility issues.

That one should ideally be `Int` too, we can separately deprecate that variant of `execute` and add one using `Int`. 

I'd much rather we course correct and add appropriate APIs rather than being consistent with past mistakes.

glbrntt (approved):

Sources/CassandraClient/Statement.swift
yifan-c: What is the rationale of using `Int`, besides the preference for public API? 
Cassandra's native protocol accepts 32-bits int. From this perspective, `Int32` reflects the API constraint precisely, meanwhile `Int` can be 64-bits depending on the platform, potentially causing overflow theoretically. In reality, no one should set the page size this large.

PR #38, opened 2024-11-01
Format code similar to SwiftNIO

PeterAdams-A:
Motivation:

Common formatting is good.

Modifications:

Enable formatting CI and format code with same
config as SwiftNIO

Result:

Common code format


.swiftformat
rnro, line 1: Why is this file retained rather than simply always using the common `.swift-format` file?

PeterAdams-A: Oh blast, maybe I've pulled the wrong thing in.  Let me check.

rnro (approved):
Looks great, thanks Peter!

PR #51, opened 2025-03-31
Enable macOS CI on merge to main and daily timer

rnro:
Motivation:

* Improve test coverage
* Check test pass/fail status
* Monitor CI throughput

Modifications:

Enable macOS CI to be run on all merges to main and on a daily timer.

Result:

Improved test coverage run out-of-band at the moment so we can get a
feeling for if any changes need to be made in the repo or in the CI
pipelines to ensure timely and stable checks.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -20,3 +20,9 @@ jobs:
   cxx-interop:
     name: Cxx interop
     uses: apple/swift-nio/.github/workflows/cxx_interop.yml@main
+
+  macos-tests:
+    name: macOS tests
+    uses: apple/swift-nio/.github/workflows/macos_tests.yml@main
+    with:
+      build_scheme: swift-cassandra-client


.github/workflows/main.yml
```diff
@@ -20,3 +20,9 @@ jobs:
   cxx-interop:
     name: Cxx interop
     uses: apple/swift-nio/.github/workflows/cxx_interop.yml@main
+
+  macos-tests:
+    name: macOS tests
+    uses: apple/swift-nio/.github/workflows/macos_tests.yml@main
+    with:
+      build_scheme: swift-cassandra-client
```
glbrntt, line 28: Is there a `-Package` suffix missing here?

czechboy0: I think it omits that suffix if there's only a single product

rnro: Yeah that's my understanding. I got these by doing `xcodebuild -list` and manually picking something which looked sensible.

glbrntt: Good to know, thanks!

glbrntt (approved):

czechboy0 (approved):

PR #52, opened 2025-03-31
Partially disable macOS CI

rnro:
Partially disable macOS tests
* Disable unit tests because they don't have a running Cassandra server to test against
* Disable watchOS and tvOS builds due to issue with `libuv` on these platforms: 'posix_spawn' has been explicitly marked unavailable`


czechboy0 (approved):
Great, thank you @rnro

PR #63, opened 2025-08-14
Expose configuration option of the load balancing strategy

FranzBusch:


Sources/CassandraClient/Configuration.swift
```diff
@@ -51,6 +51,48 @@ extension CassandraClient {
...
+                }
+            }
+
+            var backing: Backing
+
+            /// Returns a new round robin load balancing strategy.
+            public static func roundRobin(_ roundRobin: RoundRobin) -> Self {
+                .init(backing: .roundRobin(roundRobin))
+            }
+
```
czechboy0, line 88:
```suggestion

            public static var roundRobin: Self { .roundRobin(.init()) }

```

For convenience, since there is no additional round robin config.

czechboy0 (approved):
One suggestion, otherwise lgtm

PR #65, opened 2025-09-22
Enable Swift 6.2 jobs in CI

rnro:
Motivation:

Swift 6.2 has been released, we should add it to our CI coverage.

Modifications:

Add additional Swift 6.2 jobs where appropriate in main.yml, pull_request.yml

Result:

Improved test coverage.


FranzBusch (approved):

rnro:
This will need an admin merge because it uses parameters which are only added in this PR. I can split it out If needed but I hoped to avoid the churn.

PR #67, opened 2025-10-03
Add map data type support for Cassandra

ericraio:
Motivation:

The Cassandra client previously supported scalar types and arrays, but lacked support for map types which are a fundamental Cassandra collection type. Maps are commonly used for storing key-value pairs in Cassandra schemas (e.g., user preferences, metadata, configuration settings). Without map support, users had to work around this limitation or couldn't use certain Cassandra features effectively.

Modifications:

- Created Statement+Maps.swift with map binding logic for all 63 type combinations (7 key types × 9 value types)
- Added all 63 map enum cases to Statement.Value enum
- Implemented bindMap<K,V> helper using CASS_COLLECTION_TYPE_MAP
- Created Data+Maps.swift with map reading logic for all 63 combinations
- Implemented toMap<K,V> using cass_iterator_from_map(), cass_iterator_get_map_key(), and cass_iterator_get_map_value()
- Added 63 map properties on Column and 126 convenience methods on Row
- Extended testMapTypes to comprehensively test all 63 map combinations
- Organized map-related code in separate extension files to keep codebase clean

Supported key types: Int8, Int16, Int32, Int64, String, UUID, TimeBasedUUID
Supported value types: Int8, Int16, Int32, Int64, Float32, Double, Bool, String, UUID

Result:

Users can now bind and read all Cassandra map types with full type safety. The implementation follows the same pattern as arrays, providing consistent API ergonomics. All 63 map type combinations are tested and verified to work correctly with round-trip insert/read operations.

### DIFF
--- a/Sources/CassandraClient/Statement.swift
+++ b/Sources/CassandraClient/Statement.swift
@@ -94,6 +94,8 @@ extension CassandraClient {
                     result = try self.bindArray(array, at: index)
                 case .stringArray(let array):
                     result = try self.bindArray(array, at: index)
+                default:
+                    result = try self.bindMapCases(parameter, index)
                 }
 
                 guard result == CASS_OK else {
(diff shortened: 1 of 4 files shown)


Sources/CassandraClient/Statement.swift
```diff
@@ -201,6 +203,76 @@ extension CassandraClient {
             case float32Array([Float32])
             case doubleArray([Double])
             case stringArray([String])
+
+            case int8Int8Map([Int8: Int8])
```
FranzBusch, line 207: This is an API breaking which is okay since we are pre 1.0.0 but we should probably come up with a way to introduce new types without breaking API. This doesn't need to be solved in this PR though

PR #87, opened 2026-04-11
Add column encryption support for Swift Cassandra client

shruti-p-s:
Add column-level encryption for Swift Cassandra client

Motivation:
Sensitive data stored in Cassandra (e.g. SSNs, credit card numbers) needs to be encrypted at the column level so that plaintext is never exposed in the database. The solution needs to support key rotation without requiring all data to be re-encrypted at once.

Modifications:
Core encryption engine: Encryptor using AES-256-GCM with per-column and per-row key derivation. Self-describing envelope format.
Write path + manual read: Encrypted parameter types for writes, decryption convenience methods for reads, encryptor wiring through Configuration → Session → Statement.
Automatic decryption of Encrypted<T> struct fields via the existing Codable query path.

Result:
Users can encrypt columns on write, decrypt on read (manually or automatically via Codable), and rotate keys without downtime.

### DIFF
--- a/Sources/CassandraClient/Encryptor.swift
+++ b/Sources/CassandraClient/Encryptor.swift
@@ -108,7 +108,7 @@ extension CassandraClient {
     ///
     /// This type cannot be constructed directly — use `encodeKeyComponents` to
     /// ensure key bytes are always length-prefixed and unambiguous.
-    public struct PrimaryKey: Sendable {
+    public struct PrimaryKey: Sendable, Equatable {
         internal let data: Data
     }
 }
--- a/Sources/CassandraClient/Encryptor.swift
+++ b/Sources/CassandraClient/Encryptor.swift
@@ -1,0 +1,412 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift Cassandra Client open source project
+//
+// Copyright (c) 2022-2025 Apple Inc. and the Swift Cassandra Client project authors
+// Licensed under Apache License v2.0
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of Swift Cassandra Client project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import Crypto
+import Foundation
+
+// MARK: - EncryptionContext
+
+extension CassandraClient {
+    /// Context identifying the encrypted column and row for key derivation.
+    ///
+    /// Renaming keyspace, table, or column after data has been encrypted
+    /// will make that data permanently unreadable.
+    public struct EncryptionContext {
+        public let keyspace: String
+        public let table: String
+        public let column: String
+        /// Full primary key bytes (partition key + clustering columns).
+        public let primaryKey: Data
+
+        public init(keyspace: String, table: String, column: String, primaryKey: Data) {
+            self.keyspace = keyspace
+            self.table = table
+            self.column = column
+            self.primaryKey = primaryKey
+        }
+
+        internal var contextString: String { "\(self.keyspace).\(self.table).\(self.column)" }
+
+        public func withColumn(_ column: String) -> EncryptionContext {
+            EncryptionContext(keyspace: self.keyspace, table: self.table, column: column, primaryKey: self.primaryKey)
+        }
+
+        public func withPrimaryKey(_ primaryKey: Data) -> EncryptionContext {
+            EncryptionContext(keyspace: self.keyspace, table: self.table, column: self.column, primaryKey: primaryKey)
+        }
+    }
+
+    /// Partial encryption context supplied per-row during Codable decoding.
+    /// The `column` field is derived automatically from the struct's property name.
+    public struct RowEncryptionContext {
+        public let keyspace: String
+        public let table: String
+        public let primaryKey: Data
+
+        public init(keyspace: String, table: String, primaryKey: Data) {
+            self.keyspace = keyspace
+            self.table = table
+            self.primaryKey = primaryKey
+        }
+
+        internal func forColumn(_ column: String) -> EncryptionContext {
+            EncryptionContext(keyspace: self.keyspace, table: self.table, column: column, primaryKey: self.primaryKey)
+        }
+    }
+}
+
+extension CassandraClient.EncryptionContext: Sendable {}
+
+// MARK: - Encrypted wrapper
+
+extension CassandraClient {
+    /// Wrapper that marks a value for transparent encryption on the write path
+    /// and decryption on the read path via Codable.
+    public struct Encrypted<T>: Codable where T: Codable {
+        public let value: T
+        public init(_ value: T) { self.value = value }
+
+        /// Always throws — decoding is intercepted by `RowDecodingContainer` before this is called.
+        /// This conformance exists so `Encrypted<T>` satisfies the `Codable` requirement on struct fields.
+        /// Using `Encrypted<T>` with a standard `Decoder` (e.g. `JSONDecoder`) is not supported.
+        public init(from decoder: Decoder) throws {
+            throw CassandraClient.Error.decryptionError(
+                "Encrypted<T> should be decoded via RowDecoder, not directly"
+            )
+        }
+
+        public func encode(to encoder: Encoder) throws {
+            throw CassandraClient.Error.encryptionError(
+                "Encrypted<T> cannot be encoded directly. Use Statement.Value.encryptedXxx() on the write path."
+            )
+        }
+    }
+}
+
+// Only Sendable when T is Sendable
+extension CassandraClient.Encrypted: Sendable where T: Sendable {}
+
+// MARK: - KeysHolder
+
+extension CassandraClient {
+    /// Internal cache for derived column keys and storage for master key material.
+    /// Not thread-safe on its own — the `Encryptor` that owns it must hold a lock.
+    @available(macOS 11.0, *)
+    struct KeysHolder {
+        private var masterKeys: [String: Data]
+        private var kekCache: [String: SymmetricKey] = [:]
+        private let salt = Data("swift-cassandra-client-v1".utf8)
+
+        init(masterKeys: [String: Data]) {
+            self.masterKeys = masterKeys
+        }
+
+        func hasKey(_ name: String) -> Bool {
+            self.masterKeys[name] != nil
+        }
+
+        var allKeys: [String: Data] {
+            self.masterKeys
+        }
+
+        mutating func addKey(name: String, secret: Data) {
+            self.masterKeys[name] = secret
+        }
+
+        /// Returns cached column key (KEK) or derives and caches a new one.
+        /// Cache key is "keyName:contextString" (e.g. "key-2025:prod.users.ssn").
+        mutating func deriveKEK(keyName: String, context: String) throws -> SymmetricKey {
+            let cacheKey = "\(keyName):\(context)"
+            if let cached = self.kekCache[cacheKey] {
+                return cached
+            }
+            guard let masterKeyData = self.masterKeys[keyName] else {
+                throw CassandraClient.Error.keyNotFound("Key '\(keyName)' not found in keyMap")
+            }
+            let masterKey = SymmetricKey(data: masterKeyData)
+            let kek = HKDF<SHA256>.deriveKey(
+                inputKeyMaterial: masterKey,
+                salt: self.salt,
+                info: Data(context.utf8),
+                outputByteCount: 32
+            )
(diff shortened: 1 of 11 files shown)


Sources/CassandraClient/Encryptor.swift
```diff
@@ -0,0 +1,412 @@
...
+    public final class Encryptor {
+        static let envelopeVersion: UInt8 = 0x02
+        /// AES-GCM with a per-row DEK derived deterministically via HKDF from the column KEK and primary key.
+        /// Encryption is non-deterministic (random nonce per call); only the DEK derivation is deterministic.
+        static let algorithmHKDFDerivedAESGCM: UInt8 = 0x02
+        static let nonceSize = 12
+        static let keySize = 32
+        static let maxKeyNameLength = 255
+
+        private let lock = NSLock()
```
FranzBusch: Can you use a `Mutex` instead here?

Sources/CassandraClient/Encryptor.swift
```diff
@@ -0,0 +1,412 @@
...
+                throw CassandraClient.Error.encryptionConfigError(
+                    "Key name '\(name)' contains invalid characters (allowed: alphanumeric, '-', '_')"
+                )
+            }
+        }
+    }
+}
+
+@available(macOS 11.0, *)
+extension CassandraClient.Encryptor: @unchecked Sendable {}
```
FranzBusch: If you change to a `Mutex` this is no longer required to be `@unchecked`

Sources/CassandraClient/Encryptor.swift
shruti-p-s: Done, this bumps the encryption availability to macOS 15.0 / iOS 18.0 since Mutex needs the Synchronization module.

Sources/CassandraClient/Encryptor.swift
```diff
@@ -0,0 +1,412 @@
...
+        public init(keyspace: String, table: String, column: String, primaryKey: Data) {
+            self.keyspace = keyspace
+            self.table = table
+            self.column = column
+            self.primaryKey = primaryKey
+        }
+
+        internal var contextString: String { "\(self.keyspace).\(self.table).\(self.column)" }
+
+        public func withColumn(_ column: String) -> EncryptionContext {
```
hamzahrmalik: i think the name isnt quite right here, because functions named withX usually give access to X, eg withCheckedContinuation or withUnsafeMutablePointer

I would propose somehting like 
```swift
public func copy(column: String? = nil, primaryKey: Data? = nil) -> Self {
    var s = self
    if let column {
        s.column = column
    }
    if let primaryKey {
        s.primaryKey = primaryKey
    }
    return s
}
```

However, even this is not ideal because as seen in the tests we often end up with a pattern where the base is not really used, we always call withColumn to get different versions to pass into each parameter

e.g.
```swift
.encryptedString(CassandraClient.Encrypted(name), context: baseContext),
.encryptedInt32(CassandraClient.Encrypted(age), context: baseContext.withColumn("enc_age")),
.encryptedInt64(CassandraClient.Encrypted(count), context: baseContext.withColumn("enc_count")),
```

this makes it easy to forget to add the right column name

perhaps we could instead have a separate struct which represents a partial context (without the column name) and force people to call some function on this object, to return a proper context with a column name

Sources/CassandraClient/Encryptor.swift
```diff
@@ -0,0 +1,412 @@
...
+    /// Context identifying the encrypted column and row for key derivation.
+    ///
+    /// Renaming keyspace, table, or column after data has been encrypted
+    /// will make that data permanently unreadable.
+    public struct EncryptionContext {
+        public let keyspace: String
+        public let table: String
+        public let column: String
+        /// Full primary key bytes (partition key + clustering columns).
+        public let primaryKey: Data
```
pahfat: What value should this be when the primaryKey contains more than 1 column, how should we concatenate columns, or does it not matter as long as we provide the same value for encrypting and decrypting? Maybe we could have a helper that takes in several column raw values (string, uuid, ..) and computes this primaryKey Data for us? This way users won't need to handle to conversion to Data

Sources/CassandraClient/Encryptor.swift
```diff
@@ -0,0 +1,412 @@
...
+            return kek
+        }
+
+        /// Derive a per-row data encryption key (DEK) from the column key and primary key bytes.
+        /// Not cached — each row has a unique primary key, so caching would grow unbounded.
+        mutating func deriveDEK(keyName: String, context: String, primaryKey: Data) throws -> SymmetricKey {
+            let kek = try self.deriveKEK(keyName: keyName, context: context)
+            return HKDF<SHA256>.deriveKey(
+                inputKeyMaterial: kek,
+                salt: Data(),
```
pahfat, line 226: Just curious: why is salt empty here, but not in kek derivation?

Sources/CassandraClient/Statement.swift
```diff
@@ -80,6 +90,36 @@ extension CassandraClient {
                         buffer.baseAddress,
                         buffer.count
                     )
+                case .encryptedString(let wrapped, let context):
```
pahfat, line 93: How does encrypting partition keys work?
`.encryptedString()` will encrypt some data based on the primary key, which provides a unique key per row. Should there be a way to encrypt values with a derived key that is unique per column, which we can use for encrypting partition keys?

pahfat (commented):
Thank you for the PR! My main comment is about handling encryption of partition keys

Sources/CassandraClient/Encryptor.swift
shruti-p-s: Makes sense. I updated this to have a base context and a complete one after filling in the column name.

Sources/CassandraClient/Encryptor.swift
shruti-p-s: Good point, I added a length based encoding to concatenate primary keys. For uniformity, it's better to provide the full concatenated primary key, though it doesn't actually matter as long as it is the same value for encryption and decryption.

Sources/CassandraClient/Statement.swift
shruti-p-s: I'll be handling primary key encryption  in the following PR

Sources/CassandraClient/Encryptor.swift
shruti-p-s: I'd added salt for KEK derivation so that if the same root key was used by another system, the derived keys would still be different. The DEK didn't need one because its input is already the derived KEK, which used this salt. That said, I removed the KEK salt too because it ties key derivation to this specific library.

Sources/CassandraClient/Encryptor.swift
pahfat: Thank you! Nice, I wonder if we could even use a type for PrimaryKey that we can build only from your helper function, so we don't leak the Data type to the user at all?

Sources/CassandraClient/Encryptor.swift
pahfat: Sounds good, thanks

Sources/CassandraClient/Statement.swift
pahfat: Sounds good!

pahfat (commented):
Thanks for the changes, one last nit

Sources/CassandraClient/Encryptor.swift
shruti-p-s: Ah, good idea. That way, we can also ensure uniformity. Added this. Thanks!

Sources/CassandraClient/Encryptor.swift
```diff
@@ -0,0 +1,476 @@
...
+
+        internal var contextString: String { "\(self.keyspace).\(self.table).\(self.column)" }
+
+        /// Encode typed key components into a ``PrimaryKey``.
+        ///
+        /// Each component is serialized as `[4-byte big-endian length][value bytes]`.
+        /// This length-prefixing ensures composite keys are unambiguous — for example,
+        /// `encodeKeyComponents(.string("ab"), .string("c"))` produces different bytes
+        /// from `encodeKeyComponents(.string("a"), .string("bc"))`.
+        public static func encodeKeyComponents(_ components: CassandraClient.KeyComponent...) -> PrimaryKey {
```
pahfat: nit: I think that this can be moved into `PrimaryKey`, this way you won't need the prefix in `CassandraClient.EncryptionContext.encodeKeyComponents` when invoking it. It could also be an `init` rather than a static func. Maybe either:
```
public static func from(keyComponents components: CassandraClient.KeyComponent...) -> PrimaryKey {
```
or
```
public init(from components: CassandraClient.KeyComponent...) -> PrimaryKey {
```
or something similar?

shruti-p-s: Nice catch, I've updated it to an init now.

Sources/CassandraClient/Encryptor.swift
```diff
@@ -250,6 +246,7 @@ extension CassandraClient {
         ///     Multiple keys can be provided to support key rotation — old keys decrypt existing data,
         ///     while `currentKeyName` selects which key is used for new encryptions.
         ///   - currentKeyName: The key name to use for new encryptions. Must exist in `keyMap`.
+        ///   - salt: Optional salt for HKDF key derivation. Defaults to empty.
```
pahfat, line 262: Actually, what use case did you have in mind where the user would configure this? 
(If it's fine for it to be empty or a fixed value, maybe it's safer to hardcode it to be so, so the user can't misuse this parameter)

Tests/CassandraClientTests/EncryptionIntegrationTests.swift
```diff
@@ -95,7 +95,7 @@ final class EncryptionIntegrationTests: XCTestCase {
             keyspace: self.configuration.keyspace!,
             table: tableName,
             column: "secret",
-            primaryKey: CassandraClient.EncryptionContext.encodeKeyComponents(.string(userId))
+            primaryKey: CassandraClient.PrimaryKey(from: .string(userId))
```
pahfat: optional nit: you could use `.init(from: .string(userId))` without prefix now here and below

pahfat (approved):
Thank you for the PR!

yifan-c (approved):

Sources/CassandraClient/Encryptor.swift
shruti-p-s: This was not for a specific use case, but rather to make this more application specific. Also, hardcoding a fixed value would make it client dependent which is why we moved away from it.

Sources/CassandraClient/Encryptor.swift
```diff
@@ -0,0 +1,473 @@
...
+            self.data = result
+        }
+    }
+}
+
+// MARK: - KeyComponent
+
+extension CassandraClient {
+    /// A typed key component for use with ``PrimaryKey/init(from:)``.
+    public enum KeyComponent {
```
FranzBusch: Do we expect that there might be more key components in the future? If yes we should make this a `@nonexhaustive` enum

yifan-c: More key components are expected. However, I cannot find `@nonexhaustive` in swift. Looks like swift use `@frozen` to explicitly opt-in to exhaustive. By default, enums are non-exhaustive. 
@FranzBusch , please correct me if I am wrong.

shruti-p-s: Also, if I understand correctly, the non-exhaustive default only kicks in when library evolution is enabled (`BUILD_LIBRARY_FOR_DISTRIBUTION`). Since this package is compiled from source, enums are effectively frozen either way right now. But if we do enable library evolution, this would already be non-exhaustive by default.

FranzBusch: By default in packages enum are `@frozen` implicitly. `@nonexhaustive` exists since Swift 6.2.3. So you can either guard everything here on the compiler version or you can change the enum to a struct with static vars instead.

Sources/CassandraClient/Encryptor.swift
```diff
@@ -0,0 +1,473 @@
...
+        case string(String)
+        case uuid(Foundation.UUID)
+        case int32(Int32)
+        case int64(Int64)
+        case data(Data)
+    }
```
pahfat, line 135: Could we please add a case for `Date`?

shruti-p-s: Sure, done

Sources/CassandraClient/Encryptor.swift
shruti-p-s: Got it, changed this to a struct. Thanks!

PR #90, opened 2026-04-29
Add batch execution support

ahoppen:
Add support to batch insert or update data to Cassandra using the `cass_batch_*` api in the DataStax driver.

I would like to get some feedback on whether we should expose the batch API like this where the client needs to explicitly control the size of the batch or if we should have some mechanism to let the client add statements to the batch and swift-cassandra-driver automatically flushes a batch when some threshold is reached. The tricky part here would be defining that threshold. As far as I could tell there are limits to the maximum batch size in Cassandra but I haven’t found a good way to (a) figure out that limit for the Cassandra instance that we are connected to and (b) counting the number of bytes that are currently included in a batch to compare it against that maximum batch size limit.

### DIFF
--- a/Sources/CassandraClient/Session.swift
+++ b/Sources/CassandraClient/Session.swift
@@ -104,6 +104,27 @@ public protocol CassandraSession {
     )
         async throws -> CassandraClient.PaginatedRows
 
+    /// Execute a batch of statements.
+    ///
+    /// - Parameters:
+    ///   - configuration: Options to apply to the batch.
+    ///   - eventLoop: The `EventLoop` to use, or create a new one.
+    ///   - logger: If `nil`, the client's default `Logger` is used.
+    ///   - build: Closure that adds statements to the batch.
+    func batch(
+        configuration: CassandraClient.Batch.Configuration,
+        on eventLoop: EventLoop?,
+        logger: Logger?,
+        _ build: (inout CassandraClient.Batch) throws -> Void
+    ) -> EventLoopFuture<Void>
+
+    @available(macOS 12, iOS 15, tvOS 15, watchOS 8, *)
+    func batch(
+        configuration: CassandraClient.Batch.Configuration,
+        logger: Logger?,
+        _ build: (inout CassandraClient.Batch) async throws -> Void
+    ) async throws
+
     /// Terminate the session and free resources.
     func shutdown() throws
 
@@ -368,13 +389,12 @@ extension CassandraClient {
             }
         }
 
-        func execute(
-            statement: Statement,
+        /// Ensure the session is connected, then invoke `body` on the given event loop.
+        private func withConnection<Result>(
             on eventLoop: EventLoop?,
-            logger: Logger? = .none
-        )
-            -> EventLoopFuture<Rows>
-        {
+            logger: Logger?,
+            _ body: @escaping (EventLoop, Logger) -> EventLoopFuture<Result>
+        ) -> EventLoopFuture<Result> {
             let eventLoop = eventLoop ?? self.eventLoopGroup.next()
             let logger = logger ?? self.logger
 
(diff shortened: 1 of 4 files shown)


yifan-c:
@ahoppen , thank you for raising the PR and adding this feature!

> I would like to get some feedback on whether we should expose the batch API like this where the client needs to explicitly control the size of the batch or if we should have some mechanism to let the client add statements to the batch and swift-cassandra-driver automatically flushes a batch when some threshold is reached.

The Java driver (the most mature and widely-referenced Cassandra driver) uses fully explicit batch construction, i.e. "explicitly control the size of the batch". There is no auto-flush mechanism. This is the established precedent across Cassandra drivers. 

I'd recommend following the same pattern for swift-cassandra-client for feature parity and consistency.

Sources/CassandraClient/Session.swift
```diff
@@ -444,6 +452,23 @@ extension CassandraClient {
             )
         }
 
+        func execute(
+            batch: borrowing Batch,
```
shruti-p-s: `borrowing` here means the caller still owns the Batch, but withConnection's closure is @escaping and might run after the caller's batch is freed. I think that could be a use-after-free on batchPointer.
Would consuming work instead? Same for the async overload.

shruti-p-s:
Should batch() be on the CassandraSession protocol too? For testability or mocking

shruti-p-s (commented):
Thank you for the PR!

ahoppen:
> The Java driver (the most mature and widely-referenced Cassandra driver) uses fully explicit batch construction, i.e. "explicitly control the size of the batch". There is no auto-flush mechanism. This is the established precedent across Cassandra drivers.
>
> I'd recommend following the same pattern for swift-cassandra-client for feature parity and consistency.

Great. Thanks for the guidance!

> Should batch() be on the CassandraSession protocol too? For testability or mocking

Sounds reasonable to me. I wasn’t entirely sure what the point of `CassandraSession` was :)

Sources/CassandraClient/Session.swift
ahoppen: Good catch. Thank you. `consuming` works indeed. Also changed it to only take out the pointer within `withConnection` so this would have been a compilation error.

shruti-p-s (approved):
Thank you for the PR!

ahoppen:
@yifan-c Could I get a review on this PR?

Sources/CassandraClient/Session.swift
```diff
@@ -652,13 +729,44 @@ extension CassandraClient.Session {
...
-        async throws -> CassandraClient.PaginatedRows
-    {
+    ) async throws -> CassandraClient.PaginatedRows {
         try statement.setPagingSize(pageSize)
         return CassandraClient.PaginatedRows(session: self, statement: statement, logger: logger)
     }
 
+    @available(macOS 12, iOS 15, tvOS 15, watchOS 8, *)
+    func execute(
+        batch: borrowing CassandraClient.Batch,
```
yifan-c: should this be updated to `consuming` too?

Sources/CassandraClient/Session.swift
```diff
@@ -618,17 +685,34 @@ extension CassandraClient.Session {
...
+            return try await body(logger)
+        case .disconnected:
+            lock.unlock()
+            if eventLoopGroupContainer.managed {
+                preconditionFailure("client is disconnected")
+            }
+            throw CassandraClient.Error.disconnected
+        }
+    }
+    @available(macOS 12, iOS 15, tvOS 15, watchOS 8, *)
```
yifan-c, line 709: nit: please add an empty line after `}`

yifan-c (commented):
Thanks for the patch! Left a few comments.
Please rebase.

Sources/CassandraClient/Session.swift
ahoppen: Technically I don’t think it needs to be. Since this calls an async variant of `withConnection` where `body` is not escaping and the compiler would have also complained about an escape since we pass the entire `Batch` into the closure.

The reasons why it does make sense are for parity’s sake with the other `execute` function and so you don’t accidentally execute the same batch twice (which I suppose doesn’t make sense).

Updated it to `consuming`.

yifan-c (approved):

ahoppen:
Do you have a rough timeline when I can expect a release with this change included? I’m not sure what the release schedule for this library is.

yifan-c:
> Do you have a rough timeline when I can expect a release with this change included? I’m not sure what the release schedule for this library is.

There is no release cadence defined for this project (yet). Currently, it is on-demand; a new release is published on request. There are some inflight changes. But I can publish a new release for now. 
Once the project has reached the 1.0 mark, we should be more rigor on cutting releases.

ahoppen:
If we could generate a release with these changes then, I’d appreciate that 🙂

PR #91, opened 2026-05-01
Adding support for prepared statements in Swift driver

shruti-p-s:
Adding support for prepared statements in Swift driver.

### Motivation:

The client had no way to prepare CQL queries server-side and reuse them. Every query was parsed fresh on each execution. For encrypted workloads, this also meant write-path validation could only check one direction — encrypted values bound to wrong columns — because inline queries lack column name metadata for positional parameters.    

### Modifications:

- PreparedStatement wrapper class around the C driver's CassPrepared*, exposing parameter count and column name metadata.                                                                                                                                         - prepare() and execute(prepared:) on CassandraSession protocol, with both EventLoopFuture and async variants, plus Codable decoding overload.                                                                                                                     
- Bidirectional encryption validation using prepared statement parameter metadata — catches both plaintext→encrypted and encrypted→plaintext mismatches.
- 5 integration tests (4 ELF, 1 async) and 3 encryption integration tests using prepared statements.   

### Result:

Users can prepare queries once and execute them repeatedly with different parameters, avoiding repeated server-side parsing. Encrypted workloads get full bidirectional write-path validation when using prepared statements with column registration.

### DIFF
--- a/Sources/CassandraClient/PreparedStatement.swift
+++ b/Sources/CassandraClient/PreparedStatement.swift
@@ -1,0 +1,68 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift Cassandra Client open source project
+//
+// Copyright (c) 2022-2025 Apple Inc. and the Swift Cassandra Client project authors
+// Licensed under Apache License v2.0
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of Swift Cassandra Client project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+@_implementationOnly import CDataStaxDriver
+
+extension CassandraClient {
+    /// A server-side prepared statement that can be efficiently executed multiple times with different parameters.
+    public final class PreparedStatement: @unchecked Sendable {
+        internal let rawPointer: OpaquePointer
+        private let _parameterCount: Int
+
+        internal init(rawPointer: OpaquePointer) {
+            self.rawPointer = rawPointer
+            // Count parameters by iterating until we get an error from parameterName.
+            var count = 0
+            var namePtr: UnsafePointer<CChar>?
+            var nameLength = Int()
+            while cass_prepared_parameter_name(rawPointer, count, &namePtr, &nameLength) == CASS_OK {
+                count += 1
+            }
+            self._parameterCount = count
+        }
+
+        deinit {
+            cass_prepared_free(self.rawPointer)
+        }
+
+        /// The number of bind parameters in this prepared statement.
+        public var parameterCount: Int {
+            self._parameterCount
+        }
+
+        /// Gets the column name for the parameter at the given index.
+        ///
+        /// - Parameter index: Zero-based parameter index.
+        /// - Returns: The column name, or `nil` if the index is out of bounds.
+        public func parameterName(at index: Int) -> String? {
+            var namePtr: UnsafePointer<CChar>?
+            var nameLength = Int()
+            let result = cass_prepared_parameter_name(self.rawPointer, index, &namePtr, &nameLength)
+            guard result == CASS_OK, let namePtr = namePtr else {
+                return nil
+            }
+            let buffer = UnsafeBufferPointer(start: namePtr, count: nameLength)
+            return buffer.withMemoryRebound(to: UInt8.self) {
+                String(decoding: $0, as: UTF8.self)
+            }
+        }
+
+        /// Creates a bound statement from this prepared statement, ready for parameter binding and execution.
+        ///
+        /// The returned `OpaquePointer` is a `CassStatement*` that must be freed with `cass_statement_free`.
+        internal func bind() -> OpaquePointer {
+            cass_prepared_bind(self.rawPointer)
+        }
+    }
+}
--- a/Sources/CassandraClient/PreparedStatement.swift
+++ b/Sources/CassandraClient/PreparedStatement.swift
@@ -20,17 +20,16 @@ extension CassandraClient {
         internal let rawPointer: OpaquePointer
         private let _parameterCount: Int
 
-        /// The table name for encryption context resolution. Set at prepare time to avoid
-        /// passing `options.encryptionTable` on every execute call.
-        public var encryptionTable: String?
+        /// The table name for encryption context resolution.
+        public let encryptionTable: String?
 
-        /// PK column names (partition + clustering keys) for this statement's table.
-        /// Populated lazily on first encrypted execute to avoid repeated schema metadata lookups.
-        internal var primaryKeyColumnNames: [String]?
+        /// PK column names (partition + clustering keys), looked up once at prepare time to avoid repeated schema metadata queries.
+        internal let primaryKeyColumnNames: [String]
 
-        internal init(rawPointer: OpaquePointer) {
+        internal init(rawPointer: OpaquePointer, encryptionTable: String? = nil, primaryKeyColumnNames: [String] = []) {
             self.rawPointer = rawPointer
-            // Count parameters by iterating until we get an error from parameterName.
+            self.encryptionTable = encryptionTable
+            self.primaryKeyColumnNames = primaryKeyColumnNames
             var count = 0
             var namePtr: UnsafePointer<CChar>?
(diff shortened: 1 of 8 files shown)


Tests/CassandraClientTests/EncryptionIntegrationTests.swift
```diff
@@ -728,14 +834,14 @@ final class EncryptionIntegrationTests: XCTestCase {
 
         do {
             // Bind encrypted value to a plaintext column — should be rejected
-            try self.cassandraClient.run(
-                "insert into \(tableName) (user_id, name) values (?, ?)",
+            _ = try await self.cassandraClient.execute(
+                prepared: prepared,
                 parameters: [
                     .string("user-1"),
                     .encryptedString(CassandraClient.Encrypted("oops"), context: baseContext.forColumn("name")),
```
pahfat, line 841: Just thinking out loud, I see in `validateEncryptionBindings` that we retrieve the column name from the prepared statement. I wonder if it would be possible to use that retrieved column name from the statement to automatically infer the value of `context: baseContext.forColumn("name")`, so that we no longer need to pass that context manually e.g. here. 
IIUC most of the information in this context is contained in the prepared statement: keyspace, table name and column name for each parameter of the statement. However, it might be challenging to retrieve the values in that statement that are already bound, so if a primary key is already bound, it's hard for the library to retrieve it and thus construct the encryption context

Sources/CassandraClient/PreparedStatement.swift
```diff
@@ -0,0 +1,68 @@
...
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+@_implementationOnly import CDataStaxDriver
+
+extension CassandraClient {
+    /// A server-side prepared statement that can be efficiently executed multiple times with different parameters.
+    public final class PreparedStatement: @unchecked Sendable {
```
pahfat, line 19: Just a question (not yet a request): is there a way to make a batch of prepared statements?

pahfat (commented):
Thank you for the PR! Looks good to me, I just left some thoughts about whether it would be possible to not require the user to manually provide the encryption context on every query (I feel like this would improve the API), it might not be possible but prepared statements look promising, please let me know what you think

Sources/CassandraClient/PreparedStatement.swift
yifan-c: There is no batch query to **prepare** many statements at once. 
There is batch query to **execute** many  prepared statements.

Sources/CassandraClient/Session.swift
```diff
@@ -634,6 +634,145 @@ extension CassandraClient {
             }
         }
 
+
+        /// Resolve encryption contexts for parameters that have `context: nil` using driver schema metadata.
+        /// Discovers PK columns from Cassandra's metadata cache rather than requiring EncryptionSchema registration.
+        @available(macOS 15.0, iOS 18.0, visionOS 2.0, *)
+        private func resolveEncryptionContexts(
```
pahfat, line 736: That looks great!

Sources/CassandraClient/Session.swift
```diff
@@ -634,6 +634,145 @@ extension CassandraClient {
...
+            prepared: CassandraClient.PreparedStatement,
+            parameters: [CassandraClient.Statement.Value],
+            options: CassandraClient.Statement.Options
+        ) throws -> [CassandraClient.Statement.Value] {
+            guard let tableName = options.encryptionTable else { return parameters }
```
pahfat: I wonder if this `tableName` could be stored as a property in the `PreparedStatement`? When we prepare the statement, we would also store the table name explicitly, so that when using the statement, we don't need to provide it manually which may be prone to user errors.

pahfat (approved):
Thank you for the PR, it looks great to me, I just left a comment but it's not blocking the PR. It would be good to get a review from the Swift server side

Sources/CassandraClient/Session.swift
shruti-p-s: That's a good idea! Done

Sources/CassandraClient/Session.swift
```diff
@@ -473,6 +624,202 @@ extension CassandraClient {
...
+                }
+            }
+
+            // Build a map of parameter name → index for the prepared statement.
+            var paramIndexByName: [String: Int] = [:]
+            for i in 0..<parameters.count {
+                if let name = prepared.parameterName(at: i) {
+                    paramIndexByName[name] = i
+                }
+            }
```
pahfat: Question: I noticed that all those invocations to the Cassandra driver that lead to building `pkColumnNames` should return the same information for a given PreparedStatement as it's on the same table. For that reason, rather than computing this every time we execute the prepared statement, I wonder if we could have the option to run those once just after creating the PreparedStatement and store them onto that PreparedStatement? Not sure if that's feasible, what do you think?

Tests/CassandraClientTests/EncryptionIntegrationTests.swift
```diff
@@ -728,24 +834,633 @@ final class EncryptionIntegrationTests: XCTestCase {
...
+        self.recreateClient()
+
+        let insertStmt = try await self.cassandraClient.prepare(
+            "insert into \(tableName) (partition_id, cluster_id, secret) values (?, ?, ?)"
+        )
+
+        var options = CassandraClient.Statement.Options()
+        options.encryptionTable = tableName
+
+        try await self.cassandraClient.batch { batch in
```
pahfat, line 1374: Nice. Could we maybe have a test that mixes different preparedStatements for different tables and with some that don't require encryption context resolution?

Sources/CassandraClient/Session.swift
```diff
@@ -473,6 +624,202 @@ extension CassandraClient {
...
+
+        /// Resolve encryption contexts for parameters that have `context: nil` using driver schema metadata.
+        /// Discovers PK columns from Cassandra's metadata cache rather than requiring EncryptionSchema registration.
+        @available(macOS 15.0, iOS 18.0, visionOS 2.0, *)
+        private func resolveEncryptionContexts(
+            prepared: CassandraClient.PreparedStatement,
+            parameters: [CassandraClient.Statement.Value],
+            options: CassandraClient.Statement.Options
+        ) throws -> [CassandraClient.Statement.Value] {
+            let tableName = options.encryptionTable ?? prepared.encryptionTable
```
pahfat, line 741: Nice, thank you!

pahfat (commented):
Thank you for the update! I had a think about `resolveEncryptionContexts` and left a question in the comments

Tests/CassandraClientTests/EncryptionIntegrationTests.swift
shruti-p-s: Nice catch, done!

Sources/CassandraClient/Session.swift
shruti-p-s: That's a good idea, just did it.

Sources/CassandraClient/PreparedStatement.swift
```diff
@@ -0,0 +1,76 @@
...
+        internal let rawPointer: OpaquePointer
+        private let _parameterCount: Int
+
+        /// The table name for encryption context resolution. Set at prepare time to avoid
+        /// passing `options.encryptionTable` on every execute call.
+        public var encryptionTable: String?
+
+        /// PK column names (partition + clustering keys) for this statement's table.
+        /// Populated lazily on first encrypted execute to avoid repeated schema metadata lookups.
+        internal var primaryKeyColumnNames: [String]?
```
pahfat: Those `var`s in a `class` are not thread safe. Options include making them `let` or protecting them with a lock. I feel like the first one is preferred as we might not want to go through a lock on every execution (it would also be good to avoid `@unchecked Sendable` on `PreparedStatement`, that would have helped catch this)

pahfat (changes requested):
Thank you for the improvement! I left a comment regarding thread safety

Sources/CassandraClient/PreparedStatement.swift
shruti-p-s: I see, makes sense, I updated them to `let` now, thank you!

yifan-c (approved):

PR #93, opened 2026-05-13
Avoid semaphores in async code

hamzahrmalik:
Add an async variant of the shutdown function, because the current one uses semaphores to block until shutdown has completed
The new one avoids using semaphores and is `async`. The new one is safe to call from async contexts

### DIFF
--- a/Sources/CassandraClient/Session.swift
+++ b/Sources/CassandraClient/Session.swift
@@ -200,10 +200,31 @@ import NIOCore  // for async-await bridge
     ) async throws
 
     /// Terminate the session and free resources.
+    @available(*, noasync, message: "Can block indefinitely, prefer shutdown()", renamed: "shutdown()")
     func shutdown() throws
+
+    /// Terminate the session and free resources.
+    @available(macOS 12, iOS 15, tvOS 15, watchOS 8, *)
+    func shutdownAsync() async throws
 
     /// Get metrics for this session.
     func getMetrics() -> CassandraMetrics
+}
+
+// This extension ensures we don't break the API by adding the shutdownAsync function
+// We should remove it before 1.0
+extension CassandraSession {
+    /// Terminate the session and free resources.
+    @available(macOS 12, iOS 15, tvOS 15, watchOS 8, *)
+    @available(
+        *,
+        deprecated,
+        message:
+            "You must implement this function to shutdown async. This default implementation calls the sync shutdown function"
+    )
+    func shutdownAsync() async throws {
+        try self.shutdown()
+    }
 }
 
 private let encryptionLogger = Logger(label: "cassandra.encryption")
@@ -497,10 +518,14 @@ final class CassFuture<T>: Sendable {
 
     func asNIOFuture(eventLoop: any EventLoop) -> EventLoopFuture<T> {
         let promise = eventLoop.makePromise(of: T.self)
+        self.complete(to: promise)
+        return promise.futureResult
+    }
+
+    func complete(to promise: EventLoopPromise<T>) {
         self.setResultCallback { result in
             promise.completeWith(result)
         }
-        return promise.futureResult
     }
 
     @available(macOS 12, iOS 15, tvOS 15, watchOS 8, *)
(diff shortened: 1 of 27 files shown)


Sources/CassandraClient/Session.swift
```diff
@@ -200,12 +200,33 @@ import NIOCore  // for async-await bridge
     ) async throws
 
     /// Terminate the session and free resources.
+    @available(*, noasync, message: "Can block indefinitely, prefer shutdown()", renamed: "shutdown()")
     func shutdown() throws
```
yifan-c: `renamed` points to the same method. I guess you probably wanted to update both `message:` and `renamed:` to mention `shutdownAsync`

Sources/CassandraClient/Session.swift
```diff
@@ -688,6 +715,7 @@ extension CassandraClient {
             }
         }
 
+        @available(*, noasync, message: "Can block indefinitely, prefer shutdown()", renamed: "shutdown()")
```
yifan-c: The message and renamed need update too.

Sources/CassandraClient/Session.swift
```diff
@@ -732,6 +762,53 @@ extension CassandraClient {
...
+                case .idle, .connecting, .connectingFuture, .disconnected:
+                    state = .disconnected
+                    return .alreadyShut
+                }
+            }
+            switch action {
+            case .alreadyShut:
+                return
+            case .waitThenMarkShut(let task):
+                try await task.task.value
```
yifan-c: can it throw and prevent state update?

hamzahrmalik: are you referring to the preconditionFialure below? This should literally never happen because state cannot move backwards

shruti-p-s: I read the question above as being about `task.task.value` throwing, not the precondition. If close() fails, this rethrows and the whole withLockedValue block is skipped — so state stays `.disconnecting` (not moving backwards, just never advancing), and deinit's `.disconnected` precondition trips later. Sync shutdown guards this by catching the close error and always setting `.disconnected`, should we do the same here?

Sources/CassandraClient/Session.swift
```diff
@@ -200,12 +200,33 @@ import NIOCore  // for async-await bridge
...
+    /// Terminate the session and free resources.
+    @available(macOS 12, iOS 15, tvOS 15, watchOS 8, *)
+    @available(
+        *,
+        deprecated,
+        message:
+            "You must implement this function to shutdown async. This default implementation calls the sync shutdown function"
+    )
+    func shutdownAsync() async throws {
+        try self.shutdown()
```
shruti-p-s: Heads up, I don't think this compiles on the library target: `shutdown()` is `noasync`, so calling it from this async default is an error under Swift 6. Since the default just forwards to the blocking shutdown, which is the thing noasync is meant to stop, maybe  we can run it off the pool (`DispatchQueue.global().async { try self.shutdown() }` + a continuation), or drop the default and make shutdownAsync() required?

hamzahrmalik: you're right, i think swift 6 has made this stricter, this repo was still on 5 when i opened the PR i think

DispatchQueue is a good idea let me try that

Sources/CassandraClient/Session.swift
hamzahrmalik: ah got it, will fix

shruti-p-s (approved):
Thank you for the fix!
Nit: `complete(to:)` is missing the `where T: Sendable` bound that asNIOFuture/await() have, so I think `completeWith(result)` will trigger a warning under Swift 6 (`T does not conform to Sendable`).

hamzahrmalik:
> Thank you for the fix! Nit: `complete(to:)` is missing the `where T: Sendable` bound that asNIOFuture/await() have, so I think `completeWith(result)` will trigger a warning under Swift 6 (`T does not conform to Sendable`).

You're right. ive completely reverted that change, it wasnt intended to be comitted anyway

yifan-c (approved):

PR #94, opened 2026-05-14
Use NIOLockedValueBox instead of Lock

hamzahrmalik:
This moves us closer to being able to mark Session as Sendable

I'm trying not to make huge changes in one go, I might refactor this into a nio-style state machine in a follow-up

### DIFF
--- a/Sources/CassandraClient/Session.swift
+++ b/Sources/CassandraClient/Session.swift
@@ -656,8 +656,7 @@ extension CassandraClient {
 
         private let configuration: Configuration
         private let logger: Logger
-        private var state = State.idle
-        private let lock = Lock()
+        private let _state = NIOLockedValueBox(State.idle)
 
         private let underlying: CassSession
 
@@ -681,7 +680,7 @@ extension CassandraClient {
         }
 
         deinit {
-            guard case .disconnected = (self.lock.withLock { self.state }) else {
+            guard case .disconnected = (self._state.withLockedValue { $0 }) else {
                 preconditionFailure(
                     "Session not shut down before the deinit. Please call session.shutdown() when no longer needed."
                 )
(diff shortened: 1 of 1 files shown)


yifan-c:
Maybe some flakiness. Retrying the failed run. 
```
/__w/swift-cassandra-client/swift-cassandra-client/Tests/CassandraClientTests/CassandraClientTests.swift:248: error: Tests.testQueryIterator : XCTAssertNoThrow failed: threw error "Write timeout: Operation timed out - received only 0 responses." - 
```

Sources/CassandraClient/Session.swift
```diff
@@ -532,16 +531,16 @@ extension CassandraClient {
...
-                try self.disconnect()
-            default:
-                break
+            try self._state.withLockedValue { state in
+                defer {
+                    state = .disconnected
+                }
+                switch state {
+                case .disconnected:
+                    try self.disconnect()
```
yifan-c: I believe we want to disconnect when the state is `connected`, instead of `disconnected`

Sources/CassandraClient/Session.swift
```diff
@@ -554,42 +553,38 @@ extension CassandraClient {
...
+                    let promise = eventLoop.makePromise(of: Result.self)
+                    if #available(macOS 12, iOS 15, tvOS 15, watchOS 8, *) {
+                        promise.completeWithTask {
+                            try await task.task.value
+                            return try await body(eventLoop, logger).get()
+                        }
+                    }
+                    return promise.futureResult
+                case .connected:
                     return body(eventLoop, logger)
```
yifan-c: Noticed that in the original handling, it unlock before return body. It is changed in the patch.


```swift
            // original code
            case .connected:
                self.lock.unlock()
                return body(eventLoop, logger)

```

hamzahrmalik: you are right, my bad
Let me revisit this patch, we probably need to extract an action like i did in the other withconnect function

yifan-c (approved):
The refactoring is logically equivalent to the old code.

PR #95, opened 2026-05-14
Make a nominal type for CassSession and CassFuture

hamzahrmalik:
To make it easier to understand where and how each pointer is used, wrap it in a type

The CassSession type is a non copyable struct to allow us to make a deinit. The CassFuture type cannot be non copyable because it has to capture self in closures, so it's class

This makes the type much easier to hold than a raw poitner, because these new types expose helpful functions

Furthermore, the new types are Sendable, unchecked but easily to validate because they have a small api

### DIFF
--- a/Sources/CassandraClient/Session.swift
+++ b/Sources/CassandraClient/Session.swift
@@ -298,6 +298,119 @@ extension CassandraSession {
     }
 }
 
+final class CassFuture<T>: Sendable {
+    /// This can be nonisolated because the docs state that a CassFuture is thread-safe.
+    /// See Sources/CDataStaxDriver/datastax-cpp-driver/topics/README.md
+    private nonisolated(unsafe) let rawPointer: OpaquePointer
+    private let mapper: @Sendable (OpaquePointer?) -> T
+
+    init(rawPointer: OpaquePointer, mapper: @escaping @Sendable (OpaquePointer?) -> T) {
+        self.rawPointer = rawPointer
+        self.mapper = mapper
+    }
+
+    func complete(to promise: EventLoopPromise<T>) {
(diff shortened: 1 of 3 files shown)


Sources/CassandraClient/Session.swift
```diff
@@ -298,6 +298,119 @@ extension CassandraSession {
...
+        completion: @escaping (Result<T, any Error>) -> Void
+    ) {
+        let closure = unmanagedRetainedClosure {
+            DispatchQueue.global().async {
+                let result = self.result()
+                cass_future_free(self.rawPointer)
+                completion(result)
+            }
+        }
+        cass_future_set_callback(self.rawPointer, { _, data in callAndReleaseUnmanagedClosure(data!) }, closure)
```
shruti-p-s: Should we assert on the return value of `cass_future_set_callback`?
If I understand correctly, currently in CassFuture.setResultCallback, the return value is discarded. The C driver returns `CASS_ERROR_LIB_CALLBACK_ALREADY_SET` if a callback is already registered. This would mean the completion closure never fires, causing a silent hang. 
Its not an issue today as every call site uses the future only once, but it would make this more debuggable if that changes.

shruti-p-s (approved):
Thank you for the PR, it looks great. I just left a minor comment, but I'll defer to you since you have more context.

Sources/CassandraClient/Session.swift
hamzahrmalik: good idea. i've added an assertion

yifan-c:
Thank you for the patch @hamzahrmalik. Please rebase.

yifan-c (approved):

PR #96, opened 2026-06-01
Ensure prepared statement is freed when lookupPrimaryKeyColumnNames throws

hamzahrmalik:
When we create a PreparedStatement, we pass the pointer to the CassPrepared object. The PreparedStatement object then frees the pointer in the deinit. However, we might never get that far, if we throw in self.lookupPrimaryKeyColumnNames. 

In the new version, CassPrepared owns the pointer and is responsible for freeing it


shruti-p-s (approved):
Looks good to me, thank you for the fix! It'll need rebasing because `main`'s `CassSession.prepare` now returns  `CassPrepared` instead of a raw pointer.

yifan-c (approved):

PR #97, opened 2026-06-01
Conform various types to Sendable

hamzahrmalik:
Types which are already Sendable should be marked as such

This PR avoids behaviour changes, except one small change in the Configuration struct to ensure we run on the right eventloop to satisfy the compiler that we are not sending the `cluster` object, which is not Sendable

### DIFF
--- a/Sources/CassandraClient/CassandraClient.swift
+++ b/Sources/CassandraClient/CassandraClient.swift
@@ -20,7 +20,7 @@ import NIOConcurrencyHelpers
 
 /// `CassandraClient` is a wrapper around the [Datastax Cassandra C++ Driver](https://github.com/datastax/cpp-driver)
 ///  and can be used to run queries against a Cassandra database.
-public class CassandraClient: CassandraSession {
+public final class CassandraClient: CassandraSession, Sendable {
     private let eventLoopGroupContainer: EventLoopGroupContainer
     public var eventLoopGroup: EventLoopGroup {
         self.eventLoopGroupContainer.value
@@ -376,7 +376,7 @@ public class CassandraClient: CassandraSession {
     ///
     /// When `shared`, the `EventLoopGroup` is provided externally and its lifecycle will be managed by the caller.
     /// When `createNew`, the library will create a new `EventLoopGroup` and manage its lifecycle.
-    public enum EventLoopGroupProvider {
+    public enum EventLoopGroupProvider: Sendable {
         case shared(EventLoopGroup)
         case createNew
     }
--- a/Sources/CassandraClient/Batch.swift
+++ b/Sources/CassandraClient/Batch.swift
@@ -17,7 +17,7 @@ import Foundation
 
 extension CassandraClient {
     /// The type of a batch operation.
-    public struct BatchType: Equatable {
+    public struct BatchType: Sendable, Equatable {
         let rawValue: CassBatchType
 
         /// All statements are applied atomically with a write to the batch log.
@@ -110,7 +110,7 @@ extension CassandraClient {
         }
 
         /// Batch configuration options.
-        public struct Configuration {
+        public struct Configuration: Sendable {
             /// The batch type. Defaults to `.logged`.
             public var type: CassandraClient.BatchType = .logged
             /// The batch's consistency level.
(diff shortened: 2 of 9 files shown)


Sources/CassandraClient/CassandraClient.swift
```diff
@@ -20,7 +20,7 @@ import NIOConcurrencyHelpers
 
 /// `CassandraClient` is a wrapper around the [Datastax Cassandra C++ Driver](https://github.com/datastax/cpp-driver)
 ///  and can be used to run queries against a Cassandra database.
-public class CassandraClient: CassandraSession {
```
shruti-p-s, line 23: CassandraClient is marked Sendable but Session (which it holds via defaultSession) isn't marked Sendable — this compiles because of @preconcurrency on the protocol. Is the plan to mark Session as @unchecked Sendable here, or is the Sendable refactor in a follow-up?

shruti-p-s (commented):
Thanks for the PR, Hamzah. Just had one question related to Session. I'm planning to tackle the remaining types (Rows, Cluster, SSLContext, Statement, Statement.Options, Error) in a follow-up PR, just flagging in case you're already working on any of these so we don't duplicate effort.

shruti-p-s:
> Thanks for the PR, Hamzah. Just had one question related to Session. I'm planning to tackle the remaining types (Rows, Cluster, SSLContext, Statement, Statement.Options, Error) in a follow-up PR, just flagging in case you're already working on any of these so we don't duplicate effort.

Update- I just build on this branch and not all these types show in warnings, so I'll scope down my follow up accordingly

Sources/CassandraClient/CassandraClient.swift
hamzahrmalik: i thought that, since `Session:CassandraSession` and `CassandraSession:Sendable` then `Session:Sendable` is implicit. I've made it explicit now to be sure

shruti-p-s: I'm trying to understand what this gets us for v6: since we're still in .swiftLanguageMode(.v5), doesn't `Session: Sendable` only compile because the `State/ConnectionTask (_task: Any)` violations stay as warnings? Wouldn't it hard-error once we flip to v6?

hamzahrmalik: State doesn't need to be Sendable. It's in a Mutex. I don't see any warnings right now in swift 5 on Session.

shruti-p-s: Ah, you're right, no v5 warnings, my bad. I'd actually seen this as a v6 build error and wrongly assumed v5 would show it as a warning; it doesn't. Here's what I get flipping the CassandraClient to `.swiftLanguageMode(.v6)` on this commit:

  ```
  Session.swift:659: error: stored property '_state' of 'Sendable'-conforming class 'Session' contains non-Sendable type 'State'
    note: consider making enum 'State' conform to the 'Sendable' protocol
  Session.swift:1120: note: consider making struct 'ConnectionTask' conform to the 'Sendable' protocol
  ```

I think the lock doesn't cover it because NIO has:
`extension NIOLockedValueBox: @unchecked Sendable where Value: Sendable` 
so the box is only Sendable if Value is. And State isn't Sendable because its .connecting case holds a `ConnectionTask` whose _task is typed `Any`. So for v6 we'd need State/ConnectionTask made Sendable.

hamzahrmalik: Ah ok that makes sense. I don't think we actually need state to be sendable. That's a limitation of nios lock. Because fundamentally it's only accessed in one place at a time. 

We can solve this using Mutex if platform availability allows. Otherwise we can uncheck it 

Regardless,for now does this pr look ok to you? These types do need to be marked sendable no matter what, otherwise this library is very hard to use for adopters who want strict concurrency checking

shruti-p-s: Makes sense, I'll leave it to you to use Mutex or @unchecked as a follow up. Otherwise the PR looks great, thank you!

shruti-p-s (approved):

yifan-c (approved):

PR #99, opened 2026-06-03
Check state again after connecting to ensure we don't resurrect a dead session

hamzahrmalik:
currently, if session was shut down during the connect, we unconditionally set it back to state connected again, rather than keeping it disconnected

We should instead check the current state again to make sure we don't resurrect a dead connection and mark it as connected.

Added a test which would've hit the fatalError in CassandraClient/deinit without the changes in this commit


shruti-p-s (approved):
Fix looks good, nice catch! One thing after rebase: main recently added a `.disconnectingFuture` state, so we'll need to handle it.

PR #100, opened 2026-06-03
Avoid calling disconnect whilst holding the lock

hamzahrmalik:
Instead, move to a disconnecting state and then disconnect outside the lock, then take the lock again to move to disconnected state. This follows the pattern used for connecting, and also unblocks making an async version of shutdown()


yifan-c:
It is a follow-up of https://github.com/apple/swift-cassandra-client/pull/94

yifan-c (approved):

PR #103, opened 2026-06-12
Annotate closures and generics as Sendable for strict concurrency

shruti-p-s:
## Summary
  Part of the Swift 6 strict-concurrency effort (Phase 1, annotation-only). Adds `Sendable`-related annotations so these types/closures can cross concurrency boundaries; no runtime behavior change.

  ## Changes
  - query<T>/execute<T>(prepared:) (EventLoopFuture variants): constrain the decoded generic to Sendable (T: Decodable & Sendable) and mark @preconcurrency; propagate the same through CassandraClient's forwarding execute<T>. Async variants left unconstrained.
  - query(transform:): mark the transform closure @Sendable (function @preconcurrency).
  - Statement.Options: conform to Sendable; constrain encryptionContextBuilder's backing store to (any Sendable) and its closure type to @Sendable.
  - ConnectionTask: conform to Sendable, storing _task as any Sendable (was Any) — compiler-verified, unblocks Session: Sendable under v6.
  - Snapshot effectiveOptions into a let before it's captured in the result closure (avoids capturing a mutated var).

  ## Source-breaking
  - Options.encryptionContextBuilder's setter now requires a @Sendable closure (availability-gated) — the one hard break.
  - The query/execute T: Sendable constraints and transform: @Sendable are softened by @preconcurrency: non-migrated callers are unaffected; strict-concurrency callers get a warning, not an error.

  ## Verification
  - Builds clean in Swift 5 language mode (current default).
  - With the language mode flipped to v6, the targeted diagnostics are cleared; remaining errors are pre-existing items addressed in later changes.
  - Full unit + integration suite passes (89 tests).

### DIFF
--- a/Sources/CassandraClient/Session.swift
+++ b/Sources/CassandraClient/Session.swift
@@ -1149,8 +1149,9 @@ extension CassandraClient {
             self.underlying.getMetrics()
         }
 
-        fileprivate struct ConnectionTask {
-            private let _task: Any
+        fileprivate struct ConnectionTask: Sendable {
+            // `_task` only ever holds a Sendable `Task`, erased to `Any` for availability — safe to treat as Sendable.
+            nonisolated(unsafe) private let _task: Any
 
             @available(macOS 12, iOS 15, tvOS 15, watchOS 8, *)
             var task: Task<Void, Swift.Error> {
--- a/Sources/CassandraClient/Session.swift
+++ b/Sources/CassandraClient/Session.swift
@@ -311,7 +311,7 @@ extension CassandraSession {
         options: CassandraClient.Statement.Options = .init(),
         on eventLoop: EventLoop? = .none,
         logger: Logger? = .none,
-        transform: @escaping @Sendable (CassandraClient.Row) -> T?
+        transform: @escaping (CassandraClient.Row) -> T?
     ) -> EventLoopFuture<[T]> {
         self.query(query, parameters: parameters, options: options, on: eventLoop, logger: logger).map {
             rows in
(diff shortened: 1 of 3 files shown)


Sources/CassandraClient/Session.swift
```diff
@@ -1148,8 +1149,9 @@ extension CassandraClient {
             self.underlying.getMetrics()
         }
 
-        fileprivate struct ConnectionTask {
-            private let _task: Any
+        fileprivate struct ConnectionTask: Sendable {
+            // `_task` only ever holds a Sendable `Task`, erased to `Any` for availability — safe to treat as Sendable.
+            nonisolated(unsafe) private let _task: Any
```
hamzahrmalik: Can we not write this as 
```suggestion
           private let _task: any Sendable
```

To avoid the unsafe

hamzahrmalik (approved):

Sources/CassandraClient/Session.swift
```diff
@@ -1148,8 +1149,9 @@ extension CassandraClient {
...
         }
 
-        fileprivate struct ConnectionTask {
-            private let _task: Any
+        fileprivate struct ConnectionTask: Sendable {
+            // `_task` only ever holds a `Task<Void, Error>`, erased to `any Sendable` for availability.
+            private let _task: any Sendable
 
             @available(macOS 12, iOS 15, tvOS 15, watchOS 8, *)
             var task: Task<Void, Swift.Error> {
```
FranzBusch, line 1160: Just as a note we should get rid of this Task at some point. The library should not use any unstructured tasks if possible

yifan-c (approved):

PR #106, opened 2026-06-22
make Column.isNull public

hamzahrmalik:
Users should be able to check if a column is explciitly null


ericraio (approved):
I think this makes sense

yifan-c (approved):

PR #107, opened 2026-06-25
Expose wasRequestUnexecuted on CassandraClient.Error

ericraio:
#### Motivation:

When a coordinator is bootstrapping, or is shutting down (it returns an OVERLOADED error whose server message is "Server is shutting down"), it rejects the request before any execution begins. In those cases the statement had no effect and is always safe to retry on another coordinator, even when the statement is not idempotent.

CassandraClient.Error keeps its underlying code private and bridges to a generic NSError, so a caller building a retry policy has no supported way to tell these "never executed" responses apart from errors where the statement may already have been applied. The only workaround is to match on localizedDescription, which is brittle.

#### Modifications:

- Add a public computed property wasRequestUnexecuted on CassandraClient.Error.
  - It returns true for a bootstrapping node and for an overloaded response carrying the "Server is shutting down" message, and false otherwise. 

- Add a unit test covering both true cases and a range of unrelated errors.

#### Result:

Callers can drive retry decisions from a stable, documented API instead of string-matching error descriptions: treat wasRequestUnexecuted as always safe to retry regardless of statement idempotency. No behavior change for existing code.

### DIFF
--- a/Sources/CassandraClient/Errors.swift
+++ b/Sources/CassandraClient/Errors.swift
@@ -13,6 +13,7 @@
 //===----------------------------------------------------------------------===//
 
 internal import CDataStaxDriver
+import Foundation
 
 extension CassandraClient {
     /// Possible ``CassandraClient`` errors.
@@ -795,6 +796,30 @@ extension CassandraClient {
             .init(code: .other(code: code, description: description))
         }
     }
+}
+
+extension CassandraClient.Error {
+    /// The coordinator rejected the request before execution began, so the statement had no
+    /// effect and is always safe to retry on another coordinator — even when the statement is
+    /// not idempotent.
+    ///
+    /// This is true for a bootstrapping node (which never processes the request), for
+    /// `UNAVAILABLE` (the coordinator found too few replicas before contacting any of them), and
+    /// for the `OVERLOADED` response a node sends while shutting down (server message
+    /// "Server is shutting down"). A genuinely overloaded server is not covered here, since the
+    /// request may have been applied before the overload response.
+    public var wasRequestUnexecuted: Bool {
+        switch self.code {
+        case .serverBootstrapping, .serverUnavailable:
+            return true
+        case .serverOverloaded(let message):
+            return message.contains(Self.serverShuttingDownMessage)
+        default:
+            return false
+        }
+    }
+
+    private static let serverShuttingDownMessage = "Server is shutting down"
 }
 
 extension CassandraClient {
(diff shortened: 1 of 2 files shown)


Sources/CassandraClient/Errors.swift
```diff
@@ -14,6 +14,8 @@
 
 internal import CDataStaxDriver
 
+import Foundation
```
shruti-p-s, line 16: Do we need this import? `String.contains(_:)` is stdlib, and I see it compiles/runs without Foundation.

Sources/CassandraClient/Errors.swift
```diff
@@ -755,6 +757,29 @@ extension CassandraClient {
...
+extension CassandraClient.Error {
+    /// The coordinator rejected the request before execution began, so the statement had no
+    /// effect and is always safe to retry on another coordinator — even when the statement is
+    /// not idempotent.
+    ///
+    /// This is true for a bootstrapping node (which never processes the request) and for the
+    /// `OVERLOADED` response a node sends while shutting down (server message
+    /// "Server is shutting down"). A genuinely overloaded server is not covered here, since the
+    /// request may have been applied before the overload response.
+    public var wasRequestUnexecuted: Bool {
```
shruti-p-s, line 810: Should `.serverUnavailable` also map to `true` here? It looks like the cpp driver retries UNAVAILABLE without an `is_idempotent()` gate, same as IS_BOOTSTRAPPING, both are pre-execution rejections. Since we handle `.serverBootstrapping` should we also add a `.serverUnavailable` case?

shruti-p-s (commented):
Thank you for the PR, this is very helpful! I just left some minor comments

Sources/CassandraClient/Errors.swift
ericraio: I think this makes sense.

Sources/CassandraClient/Errors.swift
ericraio: Will double check but should be removable.

ericraio: Looked it over and `String.contains(_:String)` for substring search comes from Foundation's StringProtocol extension, not stdlib.

Sources/CassandraClient/Errors.swift
ericraio: Just added the serverUnavailable as well.

shruti-p-s (approved):
Looks good to me, thanks for the fix Eric!

PR #109, opened 2026-07-01
Enable Swift 6 strict concurrency on the library target

shruti-p-s:
### Motivation

  The package declares `swift-tools-version:6.1` but opts out of strict-concurrency checking via `.swiftLanguageMode(.v5)`. This is the final PR of the migration: it removes that opt-out on the library target so `CassandraClient` builds under the Swift 6 language mode with data-race checking on.

  ### Modifications

  - Remove `.swiftLanguageMode(.v5)` from the `CassandraClient` library target. 
  - `CassFuture`: `@Sendable` completion + `where T: Sendable` on the bridge methods; `syncWait` uses `NIOLockedValueBox` instead of a callback-mutated `var`.
  - `Rows: Sendable` (result pointer `nonisolated(unsafe)`, justified by the driver's read-only / concurrent-read contract); `Rows.Iterator` stays non-Sendable.
  - New internal `CassPrepared` Sendable wrapper so the prepare future carries a Sendable value.
  - Synchronous `withConnection`: `Result: Sendable` + `@Sendable` body; the non-Sendable `statement` / `batch` captures are confined.
  - `Statement` marked explicitly non-Sendable (unavailable `Sendable` conformance).
  - `execute(statement:)` and the non-paginated EventLoopFuture read API (`query`/`run`/`execute(prepared:)` and their generic variants) take `sending` arguments, closing a caller-race where one statement could be shared across two in-flight requests. The async variants are intentionally left non-`sending`.

  ### Result

  - The `CassandraClient` library compiles clean under the Swift 6 language modev(zero errors/warnings).
  - Full test suite passes (95 tests).
  - Source-breaking: callers of the non-paginated EventLoopFuture `query`/`run`/`execute`/`execute(prepared:)` now pass `sending` arguments (in addition to the `@Sendable` closure parameters introduced in earlier PRs).

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -166,9 +166,6 @@ let package = Package(
                 .product(name: "Logging", package: "swift-log"),
                 .product(name: "Crypto", package: "swift-crypto"),
                 .product(name: "Metrics", package: "swift-metrics"),
-            ],
-            swiftSettings: [
-                .swiftLanguageMode(.v5)
             ]
         ),
 
--- a/Sources/CassandraClient/CassandraClient.swift
+++ b/Sources/CassandraClient/CassandraClient.swift
@@ -120,7 +120,7 @@ public final class CassandraClient: CassandraSession, Sendable {
     ///
     /// - Returns: The resulting ``Rows``.
     public func execute(
-        statement: Statement,
+        statement: sending Statement,
         on eventLoop: EventLoop?,
         logger: Logger? = .none
     )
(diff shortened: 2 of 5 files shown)


hamzahrmalik (approved):
I'm quite sure this is safe. We will be able to remove most of the remaining `nonisolated(unsafe)` once we have https://github.com/apple/swift-nio/pull/3643

yifan-c (approved):

PR #110, opened 2026-07-04
Bump minimum deployment target to macOS 15 / iOS 18 / visionOS 2

### LINKED ISSUE #89: Bump minimum deployment target to remove @available backing store workarounds before 1.0.0
The encryption API requires macOS 15 / iOS 18 / visionOS 2.0 (CryptoKit). Since the library's minimum deployment target is lower, all encryption code is gated behind @available annotations. Some stored properties also require type-erased Any/AnyObject? backing stores since Swift doesn't support @available on stored properties.

Before 1.0.0, the minimum deployment target should be bumped to macOS 15 / iOS 18 / visionOS 2.0 to remove these workarounds.

Hardikrepo:
## Summary
- Bumps the package's platform floor to macOS 15 / iOS 18 / visionOS 2, the versions the column-encryption feature's CryptoKit APIs (HKDF, AES-GCM) actually require.
- Removes the `@available(macOS 15.0, iOS 18.0, visionOS 2.0, *)` annotations and the type-erased `(any Sendable)?` / `AnyObject?` backing stores they forced on `Configuration.encryptor`/`.encryptionSchemas`, `Statement.Options.encryptionContextBuilder`/`.encryptionTable`, and `Statement`'s internal encryptor, along with the paired `if #available { ... } else { ... }` runtime branches in `Session.swift` that could never actually take the "unavailable" branch once the deployment target guarantees the API.
- The separate `@available(macOS 12, iOS 15, tvOS 15, watchOS 8, *)` gate used for async/await APIs is unrelated to CryptoKit and is left in place, since tvOS/watchOS floors aren't being touched here.

Resolves #89.

## Test plan
- [ ] CI (`swift build` / `swift test` across the Linux/macOS matrix) — no Swift toolchain was available in the environment this was authored in to verify locally.
- [ ] Confirm the encryption test suite still passes unconditionally now that the availability guards are gone.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -96,6 +96,11 @@ do {
 
 let package = Package(
     name: "swift-cassandra-client",
+    platforms: [
+        .macOS(.v15),
+        .iOS(.v18),
+        .visionOS(.v2),
+    ],
     products: [
         .library(name: "CassandraClient", targets: ["CassandraClient"])
     ],
--- a/Sources/CassandraClient/Batch.swift
+++ b/Sources/CassandraClient/Batch.swift
@@ -92,7 +92,6 @@ extension CassandraClient {
 
         /// Add a prepared statement with parameters to this batch.
         /// Handles encryption context resolution automatically when encryption is configured.
-        @available(macOS 15.0, iOS 18.0, visionOS 2.0, *)
         public mutating func add(
             prepared: PreparedStatement,
             parameters: [Statement.Value],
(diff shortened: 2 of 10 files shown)


shruti-p-s (commented):
Thank you! I don't think we can raise the minimum deployment target though; it would drop consumers still on older OS versions, which is a breaking change for them. Could we gate the newer code paths with availability checks instead and keep the current floor? Leaning toward closing this one.

PR #111, opened 2026-07-04
Attach query text and call-site location to errors from ad-hoc queries

### LINKED ISSUE #75: Errors thrown by failed requests should contain information about the query or source location that caused them
Suppose you have a service and somewhere inside of it, too few parameters are passed to a query. This will result in an error `Error: Invalid query: Invalid amount of bind variables` to be thrown but without an attached debugger or additional logging, it’s almost impossible to tell, which query caused this issue. 

`postgres-nio`, for example, captures the `#fileID` and `#line` of the caller in the `query` method and enriches any errors with that source information, in addition to the executed query. It would be great if `swift-cassandra-client` could consider doing the same or something similar to throw more actionable errors.

Hardikrepo:
- Failures like "Invalid amount of bind variables" currently give no indication of which query or call site caused them, which is hard to track down in a service issuing many different queries -- as raised in #75, citing postgres-nio's approach of capturing `#fileID`/`#line`.
- `CassandraClient.Error` gains optional `query`/`file`/`line` fields plus an internal `annotated(query:file:line:)` that fills them in without overwriting an existing (innermost) annotation.
- Equality is now defined explicitly to compare only the error code, so existing `==` comparisons (e.g. against `.disconnected`, `.rowsExhausted`) are unaffected; `description` appends the query and `file:line` only when present, so descriptions of un-annotated errors are unchanged.
- The ad-hoc `query(...)`/`run(...)` methods on `CassandraSession` (both `EventLoopFuture`- and async-based) now take defaulted `file`/`line` parameters captured at the caller via `#fileID`/`#line`, and wrap statement creation/execution failures with the query text and that location before returning/throwing.

**Scope:** only ad-hoc `query(...)`/`run(...)` calls are covered. Prepared statement execution and batches are natural follow-ups, but `PreparedStatement` doesn't currently retain the original query text, so that needs its own change.

Resolves #75.

## Test plan
- [ ] Confirm existing error-equality and `.description`-based test assertions still pass unchanged for un-annotated errors.
- [ ] Manually verify a bind-count mismatch error now includes the query text and call site.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -96,6 +96,11 @@ do {
 
 let package = Package(
     name: "swift-cassandra-client",
+    platforms: [
+        .macOS(.v15),
+        .iOS(.v18),
+        .visionOS(.v2),
+    ],
     products: [
         .library(name: "CassandraClient", targets: ["CassandraClient"])
     ],
--- a/Sources/CassandraClient/Batch.swift
+++ b/Sources/CassandraClient/Batch.swift
@@ -92,7 +92,6 @@ extension CassandraClient {
 
         /// Add a prepared statement with parameters to this batch.
         /// Handles encryption context resolution automatically when encryption is configured.
-        @available(macOS 15.0, iOS 18.0, visionOS 2.0, *)
         public mutating func add(
             prepared: PreparedStatement,
             parameters: [Statement.Value],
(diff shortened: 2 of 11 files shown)


shruti-p-s (commented):
This is worth landing on its own, separate from the target bump. Could you rebase it onto main and keep main's `if #available { makeStatement } else  { Statement(...) }` fallbacks so it still builds on the current floor? If you can get to it by Monday EOD, great, otherwise I'm happy to open a version that does.

PR #112, opened 2026-07-04
Add a paginated query variant that decodes rows into a Decodable type

### LINKED ISSUE #31: `Session` doesn't support decoding & paging mode
`Session` supports couple of `query` types that decode the data type automatically, e.g.: https://github.com/apple/swift-cassandra-client/blob/4c395e6a0d915c40ae1d65731a0582c256fcb8da/Sources/CassandraClient/Session.swift#L133-L145

Separately, `Session` supports paging queries, for example: https://github.com/apple/swift-cassandra-client/blob/4c395e6a0d915c40ae1d65731a0582c256fcb8da/Sources/CassandraClient/Session.swift#L173-L189

What's missing is a query that'd do both paging & decoding.

Hardikrepo:
- `Session` already had `query<T: Decodable>(...)` for buffering a whole decoded result set in memory, and `query(...pageSize:)` for paging through raw `Rows`, but nothing that did both -- callers who wanted paged results decoded into a model type had to decode each `Row` by hand while iterating, as described in #31.
- Adds an async `query<T: Decodable>(...pageSize:...)` overload returning `AsyncThrowingMapSequence<PaginatedRows, T>`: it fetches and decodes one row at a time as the caller iterates, so large result sets never need to be buffered in memory just to get typed values out of them.
- Implemented as a thin `.map` over the existing `PaginatedRows` `AsyncSequence` conformance, reusing the same `makeDecoder(row:options:)` used by the non-paginated decoding query, so it picks up encryption context handling for free.
- Like the sibling `query(...)` overloads, it's distinguished purely by return type, matching the existing `query<T: Decodable & Sendable>(...)` / `query(...) -> Rows` pattern already used in this file.

Resolves #31.

## Test plan
- [ ] Add/confirm a test that iterates the new paginated-decoding sequence against a multi-page result set and checks decoded values match page-by-page.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -96,6 +96,11 @@ do {
 
 let package = Package(
     name: "swift-cassandra-client",
+    platforms: [
+        .macOS(.v15),
+        .iOS(.v18),
+        .visionOS(.v2),
+    ],
     products: [
         .library(name: "CassandraClient", targets: ["CassandraClient"])
     ],
--- a/Sources/CassandraClient/Batch.swift
+++ b/Sources/CassandraClient/Batch.swift
@@ -92,7 +92,6 @@ extension CassandraClient {
 
         /// Add a prepared statement with parameters to this batch.
         /// Handles encryption context resolution automatically when encryption is configured.
-        @available(macOS 15.0, iOS 18.0, visionOS 2.0, *)
         public mutating func add(
             prepared: PreparedStatement,
             parameters: [Statement.Value],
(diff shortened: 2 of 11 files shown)


shruti-p-s (commented):
Nice, this is just what #31 needs. Once the error-annotation PR is reworked for the current floor, this should only need a rebase onto main; the overload itself looks good. I can pick it up alongside the other one if that helps.

PR #113, opened 2026-07-04
Add withModelType: overloads for explicit decoded-type query calls

### LINKED ISSUE #20: Consider accepting model types as explicit parameters
Some feedback based on the examples in the pitch for the SSWG incubation process: for the API methods that allow returning a model object, it seems worth considering accepting the model type as an explicit parameter, for example something like:
```swift
let result = try await session.query("select * from table ...", withModelType: Model.self)
```

 While the type annotations get the job done, I think the explicit parameter is more discoverable for users, and seems more consistent with other Swift APIs. Most notably perhaps `JSONDecoder`:
```
let result = try JSONEncoder().decode(Model.self, from: myData)
```
But there are various other examples of this around the ecosystem too, e.g. [`ByteBuffer.getInteger`](https://swiftpackageindex.com/apple/swift-nio/2.46.0/documentation/niocore/bytebuffer/getinteger(at:endianness:as:)) or [`EventLoop.makePromise`](https://swiftpackageindex.com/apple/swift-nio/2.46.0/documentation/niocore/eventloop/makepromise(of:file:line:)).

We've got some similar API methods in the MongoDB driver and we went with explicit type parameters for this reason.

Hardikrepo:
- The existing `query<T: Decodable & Sendable>(...)` overloads infer `T` purely from the call site's return-type context (a variable's type annotation, or an `as [Model]` cast), which works but isn't very discoverable -- nothing at the call site itself says what type is being decoded into. #20 suggested following patterns like `JSONDecoder.decode(_:from:)` and accepting the model type as an explicit parameter instead.
- Adds sibling overloads of the non-paginated (sync and async) and the paginated-decoding `query(...)` methods that take a `withModelType model: T.Type` parameter, e.g.:
  ```swift
  session.query("select ...", withModelType: Model.self)
  ```
- These are purely additive: they're distinguished from the existing implicit-inference overloads by the presence of the `withModelType:` argument label, so nothing about the current inferred-type call pattern changes.

Resolves #20.

## Test plan
- [ ] Confirm the existing implicit-inference `query<T: Decodable>(...)` call sites still resolve to the original overloads unchanged.
- [ ] Add/confirm a test calling the new `withModelType:` overloads directly.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -96,6 +96,11 @@ do {
 
 let package = Package(
     name: "swift-cassandra-client",
+    platforms: [
+        .macOS(.v15),
+        .iOS(.v18),
+        .visionOS(.v2),
+    ],
     products: [
         .library(name: "CassandraClient", targets: ["CassandraClient"])
     ],
--- a/Sources/CassandraClient/Batch.swift
+++ b/Sources/CassandraClient/Batch.swift
@@ -92,7 +92,6 @@ extension CassandraClient {
 
         /// Add a prepared statement with parameters to this batch.
         /// Handles encryption context resolution automatically when encryption is configured.
-        @available(macOS 15.0, iOS 18.0, visionOS 2.0, *)
         public mutating func add(
             prepared: PreparedStatement,
             parameters: [Statement.Value],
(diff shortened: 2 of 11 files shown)


shruti-p-s (commented):
Nice addition. One thing I noticed: these overloads cover `query(...)` but not `execute(prepared:) -> [T]`, which also decodes rows into a model type. Since #20 talks about "the API methods that allow returning a model object" generally, was leaving `execute` out intentional, or worth adding here for parity?

PR #114, opened 2026-07-10
Add opt-in request logging (failures + slow queries)

shruti-p-s:
### Motivation

Today the library logs almost nothing about request outcomes — a `.debug` breadcrumb on execute and a `.trace` dump of parameters, but no signal when a request fails or is slow. Operators have no visibility into failures or latency through their configured `swift-log` backend.

### Modifications

- New `Configuration.requestLoggingEnabled` master switch (default `false`) gates all of the below; nothing is logged unless explicitly enabled.
- Failures are logged once, at a level chosen by a total `CassandraClient.Error` category function (`transient`/`unavailable`/ `callerFault`/`server`/`internal`).
- New `Configuration.slowQueryThresholdMillis` (default `nil`) logs successful queries whose server round-trip latency exceeds the threshold, at `.info`.
- New `Configuration.logBoundValues` (default `false`) gates bound parameter values in request logs, and the pre-existing `.trace` parameter dump — bound values are potential PII.
- Logging covers all three failure exits: request completion (execute/prepare/batch, sync and async), connect (including cluster/SSL/contact-point setup, not just the driver connect call), and preflight/construction failures (e.g. parameter binding) that throw before a request is sent.
- Prepared statements now carry their original CQL text, so prepared-path logs show the real query instead of `"(prepared)"`.
- Adds a `logger` and a `requestLoggingEnabled` requirement to the public `CassandraSession` protocol (both with default implementations, so existing conformers are unaffected) so the shared query/execute/batch implementations can log using the session's configured logger and switch.

### Result

- Adds 15 new tests (9 unit, 6 integration) covering categorization, level mapping, PII gating, the master switch, connect-failure and preflight-failure logging, and prepared/batch query text in logs.
- Full existing suite (110 tests) passes unchanged.
- No source-breaking changes: new `Configuration` fields are defaulted and the `Configuration` initializer isn't memberwise; the two new `CassandraSession` requirements ship with default implementations.

### DIFF
--- a/Sources/CassandraClient/Configuration.swift
+++ b/Sources/CassandraClient/Configuration.swift
@@ -37,11 +37,8 @@ extension CassandraClient {
         public var requestTimeoutMillis: UInt32?
         public var resolveTimeoutMillis: UInt32?
 
-        /// Master switch for request logging (failures, slow queries, connect/preflight failures). Off by default.
-        public var requestLoggingEnabled: Bool = false
-
-        /// Logs a successful query at `.info` when its latency exceeds this threshold. Only consulted when
-        /// ``requestLoggingEnabled`` is `true`. `nil` disables the check; `0` logs every success.
+        /// Logs a successful query at `.debug` when its latency reaches this threshold (ms). `nil` disables
+        /// the check; `0` logs every success.
         public var slowQueryThresholdMillis: UInt32? = nil
 
         /// Includes bound parameter values in request logs when `true`. Off by default — values are potential PII.
--- a/Sources/CassandraClient/Configuration.swift
+++ b/Sources/CassandraClient/Configuration.swift
@@ -36,6 +36,22 @@ extension CassandraClient {
         public var connectTimeoutMillis: UInt32?
         public var requestTimeoutMillis: UInt32?
         public var resolveTimeoutMillis: UInt32?
+
+        /// Master switch for request logging (failures, slow queries, connect/preflight failures). Off by default.
+        public var requestLoggingEnabled: Bool = false
+
+        /// Logs a successful query at `.info` when its latency exceeds this threshold. Only consulted when
+        /// ``requestLoggingEnabled`` is `true`. `nil` disables the check; `0` logs every success.
+        public var slowQueryThresholdMillis: UInt32? = nil
+
+        /// Includes bound parameter values in request logs when `true`. Off by default — values are potential PII.
+        public var logBoundValues: Bool = false
+
+        /// Maximum length of query text in a log record; longer text is truncated.
+        internal static let maxLoggedQueryLength = 500
+        /// Maximum length of each bound value in a log record when ``logBoundValues`` is set.
+        internal static let maxLoggedValueLength = 50
+
         public var coreConnectionsPerHost: UInt32?
         public var tcpNodelay: Bool?
         public var tcpKeepalive: Bool?
--- a/Sources/CassandraClient/RequestLogging.swift
+++ b/Sources/CassandraClient/RequestLogging.swift
@@ -1,0 +1,184 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift Cassandra Client open source project
+//
+// Copyright (c) 2022-2025 Apple Inc. and the Swift Cassandra Client project authors
+// Licensed under Apache License v2.0
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of Swift Cassandra Client project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import Dispatch
+import Logging
+import NIOCore
+
+extension CassandraClient {
+    /// Shared request-logging helper. Deliberately free of any `Session` reference so it can be unit-tested
+    /// directly with a synthesized error and a capturing `LogHandler`.
+    internal enum RequestLog {
+        /// The log level for a failure category (a mapping *from* the category, kept separate from the
+        /// categoriser so the category stays reusable by metrics).
+        static func level(for category: ErrorCategory) -> Logger.Level {
+            switch category {
+            case .transient: return .debug
+            case .unavailable, .callerFault, .server: return .warning
+            case .wrapperInternal: return .error
+            }
+        }
+
+        /// Log a failed request once, at the level for its category, with best-effort metadata.
+        /// The error field uses `shortDescription` (code label) — never `description`, which can echo the
+        /// server-provided message (potential PII).
+        static func logFailure(
+            _ error: CassandraClient.Error,
+            query: String? = nil,
+            consistency: CassandraClient.Consistency? = nil,
+            startedAt: DispatchTime? = nil,
+            boundValues: String? = nil,
+            logger: Logger
+        ) {
+            let category = error.category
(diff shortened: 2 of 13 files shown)


Sources/CassandraClient/RequestLogging.swift
```diff
@@ -0,0 +1,184 @@
...
+            query: String?,
+            consistency: CassandraClient.Consistency?,
+            thresholdMillis: UInt32?,
+            boundValues: String? = nil,
+            logger: Logger,
+            _ body: () async throws -> Value
+        ) async throws -> Value {
+            guard enabled else { return try await body() }
+            do {
+                let value = try await body()
```
hamzahrmalik: if you make a result
```swift
let result: Result<Value, any Error>
do {
    result = try await .success(body())
} catch {
    result = .failure(error)
}
```

then you can reuse the switch on line 97

Sources/CassandraClient/RequestLogging.swift
```diff
@@ -0,0 +1,184 @@
...
+            }
+        }
+
+        /// Format bound parameter values into a single string, each value capped to `maxLoggedValueLength`.
+        /// Callers pass this only when `logBoundValues` is set (values are potential PII).
+        static func formatValues(_ parameters: [CassandraClient.Statement.Value]) -> String {
+            let cap = Configuration.maxLoggedValueLength
+            return parameters.map { value in
+                let string = "\(value)"
+                return string.count > cap ? String(string.prefix(cap)) + "…" : string
```
hamzahrmalik:
```suggestion
                return string.count > cap ?string.prefix(cap) + "…" : string
```

Sources/CassandraClient/RequestLogging.swift
```diff
@@ -0,0 +1,184 @@
...
+            }.joined(separator: ", ")
+        }
+
+        private static func truncated(_ query: String) -> String {
+            let cap = Configuration.maxLoggedQueryLength
+            return query.count > cap ? String(query.prefix(cap)) + "…" : query
```
hamzahrmalik:
```suggestion
            return query.count > cap ? query.prefix(cap) + "…" : query
```

Sources/CassandraClient/RequestLogging.swift
```diff
@@ -0,0 +1,184 @@
...
+        }
+
+        /// Monotonic elapsed milliseconds (uptime clock — unaffected by wall-clock/NTP adjustments).
+        private static func elapsedMillis(since start: DispatchTime) -> UInt64 {
+            let ns = DispatchTime.now().uptimeNanoseconds &- start.uptimeNanoseconds
```
hamzahrmalik, line 186: i wonder if we should use ContinuousClock instead of DispatchTime

Sources/CassandraClient/RequestLogging.swift
```diff
@@ -0,0 +1,184 @@
...
+    /// Shared request-logging helper. Deliberately free of any `Session` reference so it can be unit-tested
+    /// directly with a synthesized error and a capturing `LogHandler`.
+    internal enum RequestLog {
+        /// The log level for a failure category (a mapping *from* the category, kept separate from the
+        /// categoriser so the category stays reusable by metrics).
+        static func level(for category: ErrorCategory) -> Logger.Level {
+            switch category {
+            case .transient: return .debug
+            case .unavailable, .callerFault, .server: return .warning
+            case .wrapperInternal: return .error
```
FranzBusch: A library should never log at `warning` or `error` level. See https://swiftpackageindex.com/apple/swift-log/1.14.0/documentation/logging/001-choosingloglevels#For-libraries

Sources/CassandraClient/RequestLogging.swift
```diff
@@ -0,0 +1,184 @@
...
+            if let consistency {
+                metadata[LogKey.consistency] = "\(consistency)"
+            }
+            if let startedAt {
+                metadata[LogKey.latencyMs] = "\(Self.elapsedMillis(since: startedAt))"
+            }
+            if let boundValues {
+                metadata[LogKey.boundValues] = "\(boundValues)"
+            }
+            logger.log(level: Self.level(for: category), "\(error.shortDescription)", metadata: metadata)
```
FranzBusch: Never interpolate something into the log message. It makes filtering for the log message in observability backends very hard. Instead pass the error to the `log` method since recent `swift-log` releases take an additional `error` parameter.

Sources/CassandraClient/Configuration.swift
```diff
@@ -36,6 +36,22 @@ extension CassandraClient {
         public var connectTimeoutMillis: UInt32?
         public var requestTimeoutMillis: UInt32?
         public var resolveTimeoutMillis: UInt32?
+
+        /// Master switch for request logging (failures, slow queries, connect/preflight failures). Off by default.
+        public var requestLoggingEnabled: Bool = false
```
FranzBusch: I would advise against property. By default the client should log everything at `debug` and `trace` level. Connection level errors are okay to surface at `info` level. Any query related error e.g. a bad index or column name should be returned to the user and at most logged at `debug` level.

You should adopt the latest task local logging interfaces of `swift-log` which will allow users to set their own logger for the scope of a request and they can then decide what the log level is going to be. cc @kukushechkin for further advice here

Sources/CassandraClient/RequestLogging.swift
shruti-p-s: Good idea, Clock/Duration would be cleaner than raw uptimeNanoseconds. But the catch is availability: ContinuousClock/Duration are macOS 13 / iOS 16+ and this timing path isn't gated. Do you want to gate it to macOS 13+ now, or keep DispatchTime and switch once the min deployment rises / the EventLoopFuture path is gone?

Sources/CassandraClient/RequestLogging.swift
shruti-p-s: Good catch, done

Sources/CassandraClient/Configuration.swift
shruti-p-s: Sounds good, the first half is done
On task-local logging, should we land it with the async/NIO migration rather than here because task-locals propagate through async/await but not through the EventLoopFuture callbacks. Happy to do it here instead if you'd rather — cc @kukushechkin, which interface should we adopt (a @TaskLocal Logger, MetadataProvider, …)?

Sources/CassandraClient/RequestLogging.swift
shruti-p-s: Ah, didn't know this, thank you. I'll fix it

Sources/CassandraClient/Configuration.swift
FranzBusch: I am fine to do the task local based logging in a follow up

kukushechkin: @shruti-p-s a `@TaskLocal` Logger, I can help with the review.

shruti-p-s: Thank you @kukushechkin!

Sources/CassandraClient/RequestLogging.swift
hamzahrmalik: oh in that case it probably makes sense to keep dispatch time then

yifan-c (approved):
Thanks! Let's have the TaskLocal logger in the follow-up.

PR #115, opened 2026-07-14
Add opt-in driver metrics poller (latency + connections)

shruti-p-s:
### Motivation:

The DataStax C++ driver aggregates per-request latency percentiles and connection counts internally and exposes them only as a pull snapshot via `getMetrics()`. Nothing bridged that snapshot into swift-metrics, so a consumer had to poll `getMetrics()` by hand and had no push-based series a dashboard could bind to. Driver-side latency/connection health was therefore invisible to the metrics backend.

### Modifications:

- Add a background poller: one NIO `RepeatedTask` on the session's `EventLoopGroup` that, each tick, reads the driver snapshot and `set`s a `Meter` per series. swift-metrics is push-only and a pre-aggregated percentile must not feed a `Timer`, so gauges (`Meter`) are the right instrument; latency units are encoded in the name (`latency_us`).
- Publish 10 latency percentiles + `connections.total`. The driver's pre-computed rates and the deprecated snapshot fields are excluded.
- Add configuration knobs, all opt-in: `metricsEnabled` (default `false`), `metricsPollIntervalMillis` (default `10000`; `nil`/`0` = poller off), and `metricsSessionName` (optional `session` dimension so multiple metrics-enabled sessions don't clobber each other's series).
- Poller lifecycle on `Session`: start on connect from the connect starter's branch only; cancel-and-await the in-flight tick before `close()`, branching on whether teardown runs on the poller's own loop.
- Test target: adopt swift-metrics' shipped `MetricsTestKit`.

### Result:

Off by default with zero cost when disabled (no task scheduled, no gauge handles created). When enabled, a background task pushes the driver's latency/connection gauges on the configured cadence. New metric names/dimensions are a de-facto public contract. No existing API or behavior changes. Verified: 3 unit + 6 integration tests pass against a live cluster, the full existing suite (104 tests) stays green, and the library builds clean under Swift 6.

### DIFF
--- a/Sources/CassandraClient/Session.swift
+++ b/Sources/CassandraClient/Session.swift
@@ -672,8 +672,16 @@ extension CassandraClient {
         private let configuration: Configuration
         private let logger: Logger
         private let _state = NIOLockedValueBox(State.idle)
+        private let _poller = NIOLockedValueBox(PollerHandle())
 
         private let underlying: CassSession
+
+        /// The running metrics poller and the event loop it ticks on. Guarded by `_poller`.
+        /// The `EventLoop` is retained so `shutdown()` can tell whether it runs on that loop.
+        private struct PollerHandle {
+            var task: RepeatedTask?
+            var loop: (any EventLoop)?
+        }
 
         private enum State {
             case idle
@@ -729,6 +737,7 @@ extension CassandraClient {
             case .wait(let future):
                 try future.wait()
             case .disconnectThenMarkShut(let promise):
+                self.stopMetricsPoller()
                 let future = self.underlying.close()
                 do {
                     try future.syncWait()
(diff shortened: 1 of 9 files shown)


Sources/CassandraClient/Session.swift
```diff
@@ -672,9 +672,17 @@ extension CassandraClient {
...
+        private let _poller = NIOLockedValueBox(PollerHandle())
 
         private let underlying: CassSession
 
+        /// The running metrics poller and the event loop it ticks on. Guarded by `_poller`.
+        /// The `EventLoop` is retained so `shutdown()` can tell whether it runs on that loop.
+        private struct PollerHandle {
+            var task: RepeatedTask?
+            var loop: (any EventLoop)?
+        }
```
FranzBusch, line 739: Since we need to migrate off NIO before the 1.0.0 release for the public APIs of this package we should probably us an `async` method and a loop with a `Task.sleep` for polling

Sources/CassandraClient/Session.swift
```diff
@@ -1368,7 +1471,15 @@ extension CassandraClient.Session {
         switch action {
         case .startedConnecting(let task):
             try await task.task.value
-            self._state.withLockedValue { $0 = .connected }
+            // Compare-and-set (see the ELF variant): guard against a concurrent `shutdown()`.
```
yifan-c: nit: let's use the full name `EventLoopFuture` instead. I believe ELF is not a common acronym. I was confused with "Executable and Linkable Format"

yifan-c (approved):
looks good.

PR #116, opened 2026-07-17
Add client-side distributed tracing for async requests

shruti-p-s:
### Motivation:
  
The async request API is instrumented for logging, but a request is invisible on a distributed trace: a caller running inside a parent span gets no child span for the Cassandra round-trip, so the query never appears on the trace timeline and its latency/errors can't be attributed to the request that caused it. 
  
  ### Modifications:
  
  - Wrap the async `execute` / `prepare` / `batch` primitives in a `.client` OpenTelemetry span, opened inside the connection body (so lazy-connect latency isn't folded in). Prepared statements and pagination are covered transitively.
  - The span nests into the caller's trace via the task-local `ServiceContext`, so there is no API or signature change. Tracing self-gates on whether a tracer is  bootstrapped (`NoOpTracer` default = near-zero cost); no configuration flag.
  - Failure spans record a sanitized status — the error code label only, never the server-provided message — plus the shared error category, so logs and traces describe a failure the same way. The original error is rethrown unchanged.
  - Adds `swift-distributed-tracing` (and transitive `swift-service-context`) as dependencies.
  - Tests: unit coverage for the attribute/error builder and consistency mapping; integration coverage (`CASSANDRA_HOST=<host>`) for span emission, parent linkage, prepared-execute CQL, batch, per-page pagination spans + parenting, and connect-fail ⇒ no span.
  
  ### Result:
  
 Applications that bootstrap a tracer get a `.client` span per async Cassandra request, nested in the caller's trace, with DB attributes and a sanitized failure status.
Applications that don't bootstrap a tracer see no behavior change (NoOp). `EventLoopFuture` callers get no spans; adopt the async API to gain them.

### DIFF
--- a/Sources/CassandraClient/RequestTracing.swift
+++ b/Sources/CassandraClient/RequestTracing.swift
@@ -1,0 +1,91 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift Cassandra Client open source project
+//
+// Copyright (c) 2022-2025 Apple Inc. and the Swift Cassandra Client project authors
+// Licensed under Apache License v2.0
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of Swift Cassandra Client project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import Tracing
+
+extension CassandraClient {
+    /// Shared request-tracing helper. Like ``RequestLog``, it holds no `Session` reference, so it is unit-testable
+    /// with an in-memory tracer and a synthesized error — no cluster.
+    internal enum RequestTrace {
+        /// Wrap an async request body in a `.client` span carrying the OpenTelemetry DB attributes, recording a
+        /// sanitized error status on throw, and rethrowing the original error unchanged.
+        ///
+        /// Uses a *manual* span (not the `withSpan` closure) deliberately: it avoids the operation-closure
+        /// `@Sendable` capture question for the non-`Sendable` `Statement`, and it avoids `withSpan`'s
+        /// auto-`recordError`, which would serialize the raw `Error.description` (the server message is PII).
+        @available(macOS 12, iOS 15, tvOS 15, watchOS 8, *)
+        static func traced<Value>(
+            _ operation: SpanName,
+            query: String?,
+            consistency: CassandraClient.Consistency?,
+            keyspace: String?,
+            _ body: () async throws -> Value
+        ) async throws -> Value {
+            var span = InstrumentationSystem.tracer.startAnySpan(operation.label, ofKind: .client)
+            defer { span.end() }
+
+            // Skip all attribute/string work when the span isn't recording (e.g. the NoOp tracer or a
+            // sampled-out span), so the disabled path stays cheap.
+            if span.isRecording {
+                span.attributes[TraceAttributeKey.system] = "cassandra"
+                span.attributes[TraceAttributeKey.operation] = operation.operation
+                if let query {
+                    span.attributes[TraceAttributeKey.queryText] = query
+                }
+                if let keyspace {
+                    span.attributes[TraceAttributeKey.namespace] = keyspace
+                }
+                if let consistency {
+                    span.attributes[TraceAttributeKey.consistencyLevel] = Self.otelToken(consistency)
+                }
+            }
+
+            do {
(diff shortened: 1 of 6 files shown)


Sources/CassandraClient/RequestTracing.swift
```diff
@@ -0,0 +1,91 @@
...
+        @available(macOS 12, iOS 15, tvOS 15, watchOS 8, *)
+        static func traced<Value>(
+            _ operation: SpanName,
+            query: String?,
+            consistency: CassandraClient.Consistency?,
+            keyspace: String?,
+            _ body: () async throws -> Value
+        ) async throws -> Value {
+            var span = InstrumentationSystem.tracer.startAnySpan(operation.label, ofKind: .client)
+            defer { span.end() }
```
FranzBusch: Any reason you use the manual `start` and `end` methods instead of `withSpan`?

Sources/CassandraClient/TraceConstants.swift
```diff
@@ -0,0 +1,57 @@
...
+    internal enum TraceAttributeKey {
+        static let system = "db.system.name"
+        static let operation = "db.operation.name"
+        static let queryText = "db.query.text"
+        static let namespace = "db.namespace"
+        static let consistencyLevel = "db.cassandra.consistency_level"
+        /// Shared failure classification — the same concept logging emits as `request.errorCategory` and
+        /// metrics tag as `errorCategory`; value is the ``CassandraClient/Error`` category (`category.rawValue`).
+        static let errorCategory = "db.cassandra.error.category"
+    }
```
FranzBusch, line 53: Wondering if we should add a dependency to https://github.com/swift-otel/swift-otel-semantic-conventions/blob/main/Sources/OTelSemanticConventions/Generated/Tracing/SpanAttributes%2Bcassandra.swift

Tests/CassandraClientTests/TracingSupport.swift
```diff
@@ -0,0 +1,183 @@
...
+    let parentSpanID: Int?
+    let attributes: [String: String]
+    let statusCode: SpanStatus.Code?
+    let statusMessage: String?
+    /// Stringified errors handed to `recordError` (should never contain the raw server message — see V2b).
+    let recordedErrors: [String]
+}
+
+/// Test span: accumulates state, and on `end()` flushes an immutable `CapturedSpan` into the tracer.
+final class TestSpan: Span, @unchecked Sendable {
```
FranzBusch: Can't you use the [InMemoryTracer](https://github.com/apple/swift-distributed-tracing/blob/main/Sources/InMemoryTracing/InMemoryTracer.swift) for testing here?

shruti-p-s: Done, you're right, I'd missed it ships.

Sources/CassandraClient/RequestTracing.swift
shruti-p-s: Yes, withSpan auto-records the thrown error, and it could contain sensitive/PII data. So to keep it off that, I'd adopted startAnySpan. But since startAnySpan is on the deprecated LegacyTracer path anyway, so I've switched to withSpan and preserved that: the closure returns a Result so the auto-record never fires.

Sources/CassandraClient/TraceConstants.swift
shruti-p-s: I wasn't sure what the project's stance on a third-party dependency was, sure.

Sources/CassandraClient/RequestTracing.swift
```diff
@@ -0,0 +1,91 @@
...
+            query: String?,
+            consistency: CassandraClient.Consistency?,
+            keyspace: String?,
+            _ body: () async throws -> Value
+        ) async throws -> Value {
+            // Use `withSpan` for span lifecycle + task-local context propagation. The operation closure
+            // returns a `Result` instead of throwing, so `withSpan`'s built-in error recording — which would
+            // serialize the raw `Error.description` (the server-provided message, PII) — never fires; we record
+            // a sanitized status ourselves and rethrow the original error to the caller.
+            let result: Result<Value, Swift.Error> = await withSpan(operation.label, ofKind: .client) { span in
```
FranzBusch, line 40: cc @kukushechkin do we have an issue tracking support for customizing how the error is recording in `withSpan`? That would be super useful for cases like this

FranzBusch (approved):

PR #117, opened 2026-07-21
Add column dimension to encryption metrics

shruti-p-s:
### Motivation:
  
The encryption metrics (cassandra.encryption.encrypt.* and cassandra.encryption.decrypt.*) were dimensioned by key name only. When several columns are encrypted under the same key, their metrics are indistinguishable — there is no way to tell which column drives encrypt or decrypt volume, latency, or decryption failures.

  ### Modifications:
  
  - Add a `column` dimension alongside the existing `keyName` on every encryption metric emission: encrypt total/duration, decrypt total/duration, and decrypt failures (including the early-parse failure path).
  - Key the cached metric handles by (keyName, column) rather than keyName alone.
  - Leave the key-rotation metric unchanged — it is not column-scoped.
  - Add a test asserting both dimensions on the encrypt, decrypt, and decrypt-failure counters.
    
Column values come from the table schema, so the added dimension stays bounded in cardinality. 

  ### Result:

Encryption metrics can be sliced by column as well as key name. The key-rotation metric is unaffected.

### DIFF
--- a/Sources/CassandraClient/EncryptionConstants.swift
+++ b/Sources/CassandraClient/EncryptionConstants.swift
@@ -23,5 +23,6 @@ extension CassandraClient {
         static let decryptFailures = "cassandra.encryption.decrypt.failures"
         static let keyRotationTotal = "cassandra.encryption.key_rotation.total"
         static let dimensionKeyName = "keyName"
+        static let dimensionColumn = "column"
     }
 }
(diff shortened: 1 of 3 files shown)


yifan-c (approved):

pahfat (approved):
Thank you!

PR #118, opened 2026-07-24
Add custom SASL authenticator

### LINKED ISSUE #16: More flexible authentication system
Hello again!

I'm attempting to use this client with [AWS Keyspaces](https://aws.amazon.com/keyspaces/). This service has two mechanisms for authentication. Right now, I'm able to use the user/pass versions, and it's working! I'm amazed!

However, the "real" way to do auth with Keyspaces is via AWS SigV4 authentication. Doing this requires building a plugin that uses  the lower level auth systems of the C++ driver. These are not currently exposed. I haven't built a system that does this yet, so I'm kind of winging it based on what I see in the (Go plugin)[https://github.com/aws/aws-sigv4-auth-cassandra-gocql-driver-plugin].

Any thoughts on how to wrap up the driver's API?

shruti-p-s:
Resolves #16

### Motivation:

The client can only authenticate with username/password today (`Configuration.username`/`password` feed `cass_cluster_set_credentials`). Systems that require a custom SASL exchange — AWS SigV4 for Amazon Keyspaces, or Kerberos — cannot be used. The DataStax C++ driver exposes the mechanism via `cass_cluster_set_authenticator_callbacks`, but nothing in the Swift layer surfaced it.

### Modifications:

- New public `CassandraClient.Authenticator` protocol (`Sendable`) with `initialResponse()`, `evaluateChallenge(_:)`, and `onSuccess(_:)`. The latter  two are defaulted, so a single-token mechanism such as SASL PLAIN implements  only `initialResponse()`.
- Bridge to the C driver: four non-capturing `@convention(c)` trampolines and a  file-scope `CassAuthenticatorCallbacks`. The authenticator travels through the  driver's `void* data` in a retained box, released by the data-cleanup callback,  so it survives `cass_cluster_free` and every reconnect. `[UInt8]` tokens are  marshaled to/from the C buffers; a thrown error is reported via  `cass_authenticator_set_error_n` so a failed exchange aborts cleanly.
- Add `Configuration.authenticator` and `Cluster.setAuthenticator(_:)`. When set,  the authenticator takes precedence over `username`/`password` in `makeCluster`.  `Configuration.description` shows `authenticator: custom`/`none` and never the  authenticator's contents.
- The single instance is shared across all connections and invoked concurrently;  the protocol doc states this and that `Sendable` does not enforce it, so a  stateful conformer must self-synchronize.
- Tests: unit coverage for authenticator byte output, the pure marshaling helpers, protocol defaults, and `description` redaction; integration coverage for connect + `onSuccess`, wrong-credential and throwing-authenticator failures, credential precedence, concurrent shared use, and box release on teardown. The  auth-dependent integration tests are gated behind `CASSANDRA_REQUIRE_AUTH` and skip otherwise.
- CI: enable `PasswordAuthenticator` on the test Cassandra, built from `.github/cassandra/Dockerfile` (the stock image has no env var for it), and set `CASSANDRA_REQUIRE_AUTH` so the auth tests run.

### Result:

Users can supply a custom SASL authenticator for mechanisms beyond username/password. Behavior is unchanged unless `authenticator` is set, and the new API is additive: the `Configuration` initializer isn't memberwise and the new field is defaulted, so existing call sites are unaffected.

### DIFF
--- a/.github/cassandra/Dockerfile
+++ b/.github/cassandra/Dockerfile
@@ -1,0 +1,14 @@
+# CI-only image: stock Cassandra 3.11 with password authentication enabled.
+#
+# The stock `cassandra:3.11` image defaults to `AllowAllAuthenticator` and exposes no environment variable
+# for the authenticator, so the only way to enable auth is to edit `cassandra.yaml`. Do it at build time
+# (a layer, always root) so the container runs the stock entrypoint unchanged. The `grep` fails the build
+# if the authenticator line is ever missing/renamed in a future base image, rather than silently booting an
+# AllowAll node — `sed` alone exits 0 on no match.
+#
+# BASE_IMAGE is overridable only so a local build can pull from a fully-qualified registry
+# (e.g. `--build-arg BASE_IMAGE=docker.io/library/cassandra:3.11`); CI uses the default.
+ARG BASE_IMAGE=cassandra:3.11
+FROM ${BASE_IMAGE}
+RUN sed -ri 's/^(authenticator:).*/\1 PasswordAuthenticator/' /etc/cassandra/cassandra.yaml \
+    && grep -q '^authenticator: PasswordAuthenticator' /etc/cassandra/cassandra.yaml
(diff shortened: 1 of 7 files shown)


yifan-c (approved):
Thanks for adding the bridge to exposing cpp driver's auth machinery.

PR #120, opened 2026-07-28
Add a paginated query variant that decodes rows into a Decodable type

### LINKED ISSUE #31: `Session` doesn't support decoding & paging mode

shruti-p-s:
### Motivation:

`Session` can decode a whole result set into `[T]` (`query<T: Decodable>`) or page through a large one as raw `Rows` (`query(...pageSize:)`), but not both — so callers paging a large result set decode each row by hand. Resolves #31.

### Modifications:

- Add an async `query<T: Decodable & SendableMetatype>(...pageSize:)` returning `AsyncThrowingMapSequence<PaginatedRows, T>` — a `.map` over the existing `PaginatedRows`  `AsyncSequence` that decodes each row via the same `makeDecoder` as the buffered decoding query (inheriting encryption-context handling), one row at a time, without materializing `[T]`.
- Cancel the `PaginatedRows` producer task on stream termination, so early exit (decode error, `break`, `prefix`) stops paging rather than fetching remaining pages.
- Integration tests: order/contents vs the buffered query, mid-stream decode failure, empty / single-page / exact-multiple boundaries, early break, and encrypted-column decrypt.

### Result:

Callers page a large result set and receive decoded values directly:

    let users: AsyncThrowingMapSequence<CassandraClient.PaginatedRows, User> =
        try await session.query("select ...", pageSize: 100)
    for try await user in users { ... }

### Known limitations:

- No aggregate "decrypted rows" debug log for paged decoding (no total without draining the stream); per-column decrypt logging and metrics still fire.
- `PaginatedRows`' page stream has no consumer backpressure (buffers unbounded), so the producer can page ahead of a slow consumer — pre-existing; a demand-driven producer is a possible follow-up.

### DIFF
--- a/Sources/CassandraClient/Data+PaginatedRows.swift
+++ b/Sources/CassandraClient/Data+PaginatedRows.swift
@@ -252,7 +252,7 @@ extension CassandraClient.PaginatedRows: AsyncSequence {
             // to overwrite previous rows so that the iterated rows differ from those
             // yielded to continuation.
             self.pageStream = AsyncThrowingStream { continuation in
-                Task {
+                let task = Task {
                     do {
                         repeat {
                             // This shouldn't throw `rowExhausted` error, but if it does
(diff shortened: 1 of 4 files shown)


Sources/CassandraClient/Data+PaginatedRows.swift
```diff
@@ -252,21 +252,24 @@ extension CassandraClient.PaginatedRows: AsyncSequence {
...
+                        } while paginatedRows.hasMorePages && !Task.isCancelled
 
                         continuation.finish()
                     } catch {
                         continuation.finish(throwing: error)
                     }
                 }
+                // Stop paging when the consumer stops early (break, `prefix`, a decode error in a
+                // downstream `map`): otherwise the producer keeps fetching pages to exhaustion.
+                continuation.onTermination = { _ in task.cancel() }
```
yifan-c, line 272: 👍

yifan-c (approved):

PR #121, opened 2026-07-29
Redact Encrypted<T> plaintext from string and reflection forms

shruti-p-s:
### Motivation:

`Encrypted<T>` and the `Statement.Value.encryptedXxx` cases had no custom string form, so Swift's default reflection rendered the wrapped plaintext verbatim. Interpolating a `Statement`, its parameters, or an `Encrypted<T>` — including the public, ungated `Statement.description` and bound-value request logs — could emit
encryption-marked plaintext in cleartext.

### Modifications:

Conform `Encrypted<T>` to `CustomStringConvertible`, `CustomDebugStringConvertible`, and `CustomReflectable`, all rendering `<redacted>`. Either string conformance covers `String(describing:)` and `String(reflecting:)`; `CustomReflectable` covers `dump(_:)`/`Mirror`. Every affected sink renders the wrapper via reflection, so one set of conformances closes all of them. Adds tests for the wrapper, all seven encrypted `Statement.Value` cases, `Statement.description`, the bound-value formatter, and `dump`/`Mirror`.

### Result:

Encrypted plaintext renders as `<redacted>` across all string and reflection forms. No behavior change to encryption, decryption, or binding.

### DIFF
--- a/Tests/CassandraClientTests/EncryptorTests.swift
+++ b/Tests/CassandraClientTests/EncryptorTests.swift
@@ -571,6 +571,86 @@ final class EncryptorTests: XCTestCase {
         }
     }
 
+    // MARK: - Plaintext redaction in string forms
+
+    /// Test plaintext that must never appear in any string form of an encrypted value.
+    private static let secret = "123-45-6789"
+
+    /// `Encrypted<T>` renders as `<redacted>` — never its wrapped plaintext — through every string form.
+    func testEncryptedWrapperRedactsPlaintext() {
+        let wrapped = CassandraClient.Encrypted(Self.secret)
+        // `==` in XCTAssertTrue, not XCTAssertEqual — the latter echoes the actual value (plaintext) on failure.
+        for rendered in ["\(wrapped)", String(describing: wrapped), String(reflecting: wrapped)] {
+            XCTAssertTrue(rendered == "<redacted>", "wrapper string form should be <redacted>")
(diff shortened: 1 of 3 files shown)


yifan-c (approved):

Tests/CassandraClientTests/EncryptorTests.swift
```diff
@@ -571,6 +571,86 @@ final class EncryptorTests: XCTestCase {
...
+                "2023-11-14"
+            ),
+        ]
+        for (label, value, plaintext) in cases {
+            // Both string forms — `String(reflecting:)` is the debug form some sinks use.
+            // Messages use only the static label — a regression must not echo the plaintext to CI.
+            for rendered in ["\(value)", String(reflecting: value)] {
+                XCTAssertTrue(rendered.contains("<redacted>"), "\(label) should render <redacted>")
+                XCTAssertFalse(rendered.contains(plaintext), "\(label) leaked plaintext")
+            }
```
yifan-c: How are the assertions different from line#584-586

shruti-p-s: The difference from 584-586 is that these exercise the enum-embedded form (`Statement.Value.encryptedString(...)`) rather than the bare wrapper. On second thought, these are little overkill. Redaction doesn't vary by payload type, so the per-type cases just repeat what 584-586 and the Statement.description test already cover. Removing.

PR #122, opened 2026-08-03
Restrict OpaquePagingStateToken to in-process use

shruti-p-s:
### Motivation:

 `Statement.setPagingStateToken(_:)` accepted any `PagingStateToken`, a public  protocol whose only requirement was `ContiguousBytes`. A consumer could conform a type over bytes from an untrusted source — an HTTP cursor handed to a client and accepted back — and have them forwarded to the driver as a paging state. A paging state encodes a server-side resume position which, per the driver's own warning, could be spoofed and potentially used to gain access to other data. The token is opaque and the driver stores it verbatim, so the library cannot validate it.
`Rows.opaquePagingStateToken()` already documented that the token is not safe to share externally. The reverse direction was open too: the token exposed its bytes publicly, which is what made serializing one out possible.

  ### Modifications:

Narrowed `setPagingStateToken(_:)` to the concrete `OpaquePagingStateToken`, which has no public initializer and can only be obtained from `Rows.opaquePagingStateToken()`, and removed the now-unreferenced `PagingStateToken` protocol.
Made `withUnsafeBytes` internal and dropped the public `ContiguousBytes`  conformance so a token's bytes are no longer part of the public API. Added `Equatable`, the only legitimate use of that accessor; `Hashable` is deliberately omitted, since the supported pattern keys by a consumer-chosen identifier and stores the token as a value.
Extended `testPagingToken` to assert that tokens from one result compare equal and tokens from different pages do not.

  ### Result:

 A token can only be obtained from a result, compared, and handed back to a statement in the same process.
 Source-breaking in two ways:
 - A consumer that passed its own `PagingStateToken` conformer, or named that protocol as a type, no longer compiles.
  - A consumer that read a token's bytes, or passed a token to a `ContiguousBytes`-taking API, no longer compiles. This is the wider of the two.

There is no replacement for resuming a query outside the process that produced it,  because a token cannot be reconstructed from bytes. Consumers that persisted a cursor externally must keep the token in memory instead.

  ### Known limitations:

  - Reflection still reaches the stored bytes, so the guarantee is that no supported API reads them, not that they are unreachable.

### DIFF
--- a/Sources/CassandraClient/Data.swift
+++ b/Sources/CassandraClient/Data.swift
@@ -48,8 +48,6 @@ extension Data {
     }
 }
 
-public protocol PagingStateToken: ContiguousBytes {}
-
 extension CassandraClient {
     /// Resulting row(s) of a Cassandra query. Data are returned all at once.
     public final class Rows: Sequence, Sendable {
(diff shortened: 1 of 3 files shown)


yifan-c (approved):
This is an API breaking change. The SemVer/Minor label is proper, given we do not have 1.0 release yet.

PR #125, opened 2026-08-11
Verify certificate chain and peer identity when SSL is enabled

shruti-p-s:
### Motivation

TLS verification asks two independent questions: is the certificate from a trusted issuer, and does it belong to the host we dialed. The driver takes these as separate bits in one mask, but `Configuration.SSL` modeled them as a pick-one enum and set one bit per case, so no value checked both. The default checked the chain and not identity, letting an attacker on the network path present any certificate from the configured issuer and read the credentials sent over it. `.peerIdentity` checked identity and not the chain, so a self-signed certificate with the right address
passed.

The new default matches the IP rather than the hostname because the driver reaches discovered peers by IP and gets their hostname from a reverse lookup it never forward-confirms, so hostname matching would rest on a PTR record an attacker can answer.

### Modifications

- `verifyFlag` defaults to `.peerIdentity`; both identity cases now also set the  chain bit.
- Removed `.default`, a no-op that read as safe but meant "no identity check". Use  `.peerCert` instead.
- `.peerIdentityDNS` without `hostnameResolution` now throws when the cluster is built.
- Both connect paths warn on `.none` and `.peerCert`, and `Configuration.description`  shows the verify mode.
- Unit tests for the mapping, the applied mask, the `hostnameResolution` rule, the warning, and the description.

### Result

Identity is verified by default. Breaking for SSL users two ways:

- Certificates without an `iPAddress` subject alternative name for the dialed address fail with `sslIdentityMismatch`. For discovered peers that is their `system.peers rpc_address`, not necessarily a configured contact point.
- `.peerIdentity` with `trustedCertificates` nil now fails chain validation. That needs trust anchors, not certificate changes; the driver loads no system trust store.

Naming `.default` is a compile error.

### DIFF
--- a/Sources/CassandraClient/Configuration.swift
+++ b/Sources/CassandraClient/Configuration.swift
@@ -242,6 +242,14 @@ extension CassandraClient {
                 try cluster.setCredentials(username: username, password: password)
             }
             if let ssl = self.ssl {
+                // The driver matches DNS identity against a hostname it only resolves when hostname
+                // resolution is on; without it peers reached by IP carry no hostname and every
+                // handshake fails the subject match.
+                if ssl.verifyFlag == .peerIdentityDNS, self.hostnameResolution != true {
+                    throw CassandraClient.Error.badParams(
+                        "SSL verifyFlag .peerIdentityDNS requires hostnameResolution to be true"
+                    )
+                }
                 try cluster.setSSL(try ssl.makeSSLContext())
             }
             if let value = self.numIOThreads {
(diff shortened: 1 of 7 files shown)


shruti-p-s:
Pre 1.0 so this change is semver-minor

Sources/CassandraClient/Configuration.swift
```diff
@@ -492,29 +520,71 @@ internal final class Cluster {
 extension CassandraClient.Configuration {
     public struct SSL: Sendable {
         public var trustedCertificates: [String]?
-        public var verifyFlag: VerifyFlag = .default
+        public var verifyFlag: VerifyFlag = .peerIdentity
```
yifan-c, line 523: Since it is a breaking change for ssl-enabled use cases, I think we need to have the document somewhere, so that the end-users who upgrade the client do not get surprised.

yifan-c (approved):

PR #126, opened 2026-08-12
Add swift-configuration support

hamzahrmalik:
Make it possible to configure Cassandra client using Swift configuration. I've followed the patterns used in async-http-client. Almost all of the configs are optional and default to the existing defaults.

None of the config is reloading, except for contact points provider.

### DIFF
--- a/Sources/CassandraClient/Configuration+SwiftConfiguration.swift
+++ b/Sources/CassandraClient/Configuration+SwiftConfiguration.swift
@@ -1,0 +1,312 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Swift Cassandra Client open source project
+//
+// Copyright (c) 2026 Apple Inc. and the Swift Cassandra Client project authors
+// Licensed under Apache License v2.0
+//
+// See LICENSE.txt for license information
+// See CONTRIBUTORS.txt for the list of Swift Cassandra Client project authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+#if compiler(>=6.2)
+import Configuration
+
+@available(macOS 15.0, iOS 18.0, watchOS 11.0, tvOS 18.0, visionOS 2.0, *)
+extension CassandraClient.Configuration {
+    /// Initializes ``CassandraClient/Configuration`` from a `ConfigReader`.
+    ///
+    /// Unless noted otherwise, each key maps onto the property of the same name, an absent key leaves that
+    /// property at its default, and every value is read once here — later provider updates do not affect the
+    /// returned configuration. `contactPoints` is the only exception.
+    ///
+    /// ## Configuration keys:
+    /// - `contactPoints` (string array, required): Initial contact points of the Cassandra cluster. Must hold
+    ///   at least one entry, none of them blank. Unlike every other key, this one is re-read from
+    ///   `configReader` on each cluster creation rather than captured, so a reloading provider's new seeds
+    ///   apply to subsequent connections. A re-read that fails validation fails that connection.
+    /// - `port` (int, optional, default: 9042): Port the cluster listens on, 1 through 65535.
+    /// - `protocolVersion` (int, optional, default: 4): Native protocol version, 1 through 5.
+    /// - `username` (string, optional): Username for plain text authentication. Unused when ``authenticator`` is set in code.
+    /// - `password` (string, optional, secret): Password for plain text authentication. Unused when ``authenticator`` is set in code.
+    /// - `keyspace` (string, optional): Keyspace the session connects to.
+    /// - `numIOThreads` (int, optional): Number of driver IO threads.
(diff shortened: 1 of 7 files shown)


Sources/CassandraClient/Configuration+SwiftConfiguration.swift
```diff
@@ -0,0 +1,312 @@
...
+    /// - `trustedCertificates` (string array, optional): PEM encoded certificates used to verify the peer.
+    /// - `verifyFlag` (string, optional, default: "default"): Verification performed on the peer's
+    ///   certificate, one of "default", "none", "peerCert", "peerIdentity" or "peerIdentityDNS".
+    /// - `cert` (string, optional): PEM encoded client certificate chain.
+    /// - `privateKey` (string, optional, secret): PEM encoded client private key.
+    /// - `privateKeyPassword` (string, secret): Password for `privateKey`. Required when `privateKey` is set.
+    ///
+    /// - Throws: If `verifyFlag` is not one of the accepted values, or `privateKey` is set without `privateKeyPassword`.
+    public init?(configReader: ConfigReader) throws {
+        guard configReader.bool(forKey: "enabled", default: false) else {
```
shruti-p-s, line 196: Should a missing `ssl.enabled` be an error when the rest of the scope is populated? A config that sets `ssl.cert` / `ssl.privateKey` but omits `enabled` connects in plaintext with no diagnostic. I see `sslIsIgnoredWhenNotEnabled` pins this, so could we keep an explicit `enabled: false` meaning off and throw only when `enabled` is absent while another `ssl.` key is set?

Sources/CassandraClient/Configuration+SwiftConfiguration.swift
```diff
@@ -0,0 +1,312 @@
...
+            return nil
+        }
+        self.init()
+
+        self.trustedCertificates = configReader.stringArray(forKey: "trustedCertificates")
+        if let verifyFlag = configReader.string(forKey: "verifyFlag", as: VerifyFlag.self) {
```
shruti-p-s: Should these throw when the value is present but unrecognized?   `string(forKey:as:)` returns `nil`  for both "absent" and "not convertible", so `verifyFlag: "peerIdentityDns"` falls back to `.default`, i.e. `CASS_SSL_VERIFY_PEER_CERT`, dropping the identity check. The `- Throws:` above says these are rejected, and `loadBalancingStrategy.strategy` already reads the `String` and throws in `default:`. Could we use that shape for `consistency`, `serialConsistency`, `prepareStrategy`  and `verifyFlag`?

Sources/CassandraClient/Configuration+SwiftConfiguration.swift
```diff
@@ -0,0 +1,312 @@
...
+
+        let rawPort = configReader.int(forKey: "port", default: 9042)
+        guard let port = Int32(exactly: rawPort), (1...65535).contains(port) else {
+            throw CassandraClient.ConfigurationError(
+                message: "'port' must be between 1 and 65535, got \(rawPort)"
+            )
+        }
+
+        let rawProtocolVersion = configReader.int(forKey: "protocolVersion", default: 4)
+        guard let protocolVersion = Int32(exactly: rawProtocolVersion).flatMap(ProtocolVersion.init(rawValue:)) else {
```
shruti-p-s: The driver only accepts V3...V4 here, and V5 needs `cass_cluster_set_use_beta_protocol_version`,  which I don't see wrapped. So `protocolVersion: 1|2|5` passes init and then throws on every `makeCluster`, and the `Valid values:` message lists them as valid. Can we range-check 3...4 here?

shruti-p-s (commented):
Just a few comments, PR looks great otherwise, thank you Hamzah!

Sources/CassandraClient/Configuration+SwiftConfiguration.swift
hamzahrmalik: if someone has set ssl to true, and has set correspinding config, and then wants to temporarily turn it off, they shouldn't have to delete the config imo, they should only need to set the enabled flag to off

shruti-p-s: Agreed, and that case would keep working. My comment wasn't clear, I only meant the case where `enabled` is missing entirely. `bool(forKey:default:)` collapses "absent" and "explicitly false" into the same `false`, but `bool(forKey:)` returns `Bool?`, so we can tell them apart. Should we throw only on the `nil` case when another `ssl.` key is set?

shruti-p-s (approved):
Thank you, the fix looks good!

Sources/CassandraClient/Configuration+SwiftConfiguration.swift
hamzahrmalik: im still not sure about throwing. what about logging a warning?

shruti-p-s: Sure, that makes sense

PR #127, opened 2026-08-20
Make Statement.setPagingSize public and validate the page size

shruti-p-s:
### Motivation:

`Statement.setPagingStateToken` is public, but `setPagingSize` was internal, so the page size could only be set through the `execute`/`query` variants that take a `pageSize` — and those build the statement themselves. Bounding the page of a caller-configured statement, which is what driving pagination by hand requires, was not possible. Supersedes #30 and keeps its commits.

### Modifications:

Made `setPagingSize` public, taking `Int` rather than `Int32`. Rejected sizes outside `1...Int32.max` with `badParams`: the driver's page size is a C `int`, so a larger value would trap on the narrowing conversion, and a non-positive one disables paging in the driver rather than limiting the page, returning the whole result set. Documented the constraint on all eight public entry points that take a `pageSize`, and qualified "**All** rows are returned" wherever a caller-supplied statement can now bound it. Added unit tests for the accepted range, both paginated entry points, and the documented overwrite of a size already set on the statement, plus an integration test that pages a statement by hand.

### Result:

A per-statement page size can be set and paired with `setPagingStateToken(_:)` to walk pages by hand, and a page size the driver would ignore is reported instead of silently changing the result shape.

Source-compatible, with one behavior change: the existing `pageSize: Int32` entry points share the validation, so a `pageSize` of `0` or below now fails where it previously returned every row from a paginated call. Unifying those parameters on `Int` is a follow-up.

Co-authored-by: Igor Milyakov <imilyakov@apple.com>

### DIFF
--- a/Tests/CassandraClientTests/CassandraClientTests.swift
+++ b/Tests/CassandraClientTests/CassandraClientTests.swift
@@ -427,7 +427,7 @@ final class Tests: XCTestCase {
         XCTAssertLessThanOrEqual(
             secondIDs.count,
             pageSize,
-            "the resumed page should be bounded the same way"
+            "the resumed page should not exceed the page size"
         )
         XCTAssertFalse(secondIDs.isEmpty, "the resumed page should not be empty")
         XCTAssertTrue(
--- a/Tests/CassandraClientTests/CassandraClientTests.swift
+++ b/Tests/CassandraClientTests/CassandraClientTests.swift
@@ -371,6 +371,71 @@ final class Tests: XCTestCase {
         XCTAssertEqual(id1.int32, id2.int32)
     }
 
+    /// A page size set on the statement itself bounds the result of a plain (non-paginated) execute,
+    /// which is what lets a caller drive paging by hand: read a page, keep its token, resume from it.
+    func testPagingSizeOnStatement() throws {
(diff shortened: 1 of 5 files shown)


Tests/CassandraClientTests/CassandraClientTests.swift
```diff
@@ -371,6 +371,71 @@ final class Tests: XCTestCase {
...
+        // Resuming from the first page's token with the same page size yields the following page.
+        let secondStatement = try CassandraClient.Statement(query: query)
+        try secondStatement.setPagingSize(pageSize)
+        try secondStatement.setPagingStateToken(try firstPage.opaquePagingStateToken())
+        let secondPage = try self.cassandraClient.execute(statement: secondStatement, on: nil).wait()
+        let secondIDs = secondPage.compactMap { $0.column(0)?.int32 }
+        XCTAssertLessThanOrEqual(
+            secondIDs.count,
+            pageSize,
+            "the resumed page should be bounded the same way"
```
yifan-c: nit: "the same way" could be omitted, as it could be hard to reason about what's the same way when the assertion fails. 
Not important, feel free to ignore.

yifan-c (approved):

PR #128, opened 2026-08-26
Allow contact points provider to be set separately from ConfigReader

hamzahrmalik:
### Motivation:

The contact points provider will often come from something whic needs to be manually wired in, rather than something which can be represented in config. For example, service discovery

The config reader inits added in #126 do not let users make the config using a ConfigReader but then override the contact points provider, because they require the config to contain the provider

### Modifications:

Create a second init which allows users to specify the contact points provider separately from the ConfigReader

### DIFF
--- a/Sources/CassandraClient/Configuration+SwiftConfiguration.swift
+++ b/Sources/CassandraClient/Configuration+SwiftConfiguration.swift
@@ -24,6 +24,10 @@ extension CassandraClient.Configuration {
     /// property at its default, and every value is read once here — later provider updates do not affect the
     /// returned configuration. `contactPoints` is the only exception.
     ///
+    /// To supply the contact points in code instead — e.g. from service discovery — use
+    /// ``init(configReader:contactPointsProvider:logger:)``, which reads every other key exactly as this
+    /// initializer does.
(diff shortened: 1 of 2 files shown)


Sources/CassandraClient/Configuration+SwiftConfiguration.swift
```diff
@@ -79,6 +83,68 @@ extension CassandraClient.Configuration {
...
+                "'contactPoints' is set but the contact points are supplied in code, so the key is ignored.",
+                metadata: [
+                    CassandraClient.ConfigurationLogKey.ignoredKeys: .string("contactPoints")
+                ]
+            )
+        }
+
+        let (port, protocolVersion) = try Self.portAndProtocolVersion(from: configReader)
+        self.init(
+            contactPointsProvider: contactPointsProvider,
```
shruti-p-s, line 134: Minor thing, the config path rejects an empty list or a blank entry (`contactPoints(from:)`) but this initializer passes the provider through unchecked, and a blank entry clears the addresses added before it. Should we add the same validation here?

shruti-p-s (approved):

PR #129, opened 2026-08-27
Require encryptedColumns and correct public interface defects ahead of 1.0

shruti-p-s:
### Motivation

`EncryptionSchema.init` defaulted `encryptedColumns` to an empty set, which left the write-path binding validation with nothing to enforce: a plaintext value bound to a column meant to be encrypted was written as cleartext with no error.

Several other parts of the public interface misreport themselves — doc comments state defaults the code contradicts, and a `private` error type reaches adopters through every public `Decodable` overload of `query` and `execute`, unnameable and uninspectable. Plain-data value types also lack `Hashable`, which is cheaper to add before 1.0 than after.

### Modifications

Remove the `encryptedColumns` default, and document what binding validation enforces on each execution path, since prepared and non-prepared execution differ.

Replace the private `DecodingError` with `Swift.DecodingError`, so decoding failures are nameable and carry a `codingPath`. The typed column accessors collapse four causes into `nil`, so probe the untyped accessor to separate them: a missing column is `keyNotFound`, a NULL read into a non-Optional property is `valueNotFound`, invalid UTF-8 in a text column is `dataCorrupted`, and an unconvertible stored type is `typeMismatch`. Decoder misconfiguration throws `encryptionConfigError` instead of masquerading as a decoding failure; ciphertext that fails authentication still throws `decryptionError` from the encryptor.

Conform `Error` and `ConfigurationError` to `LocalizedError` so `localizedDescription` reports the message. Add `Hashable` to nine value types; `Encrypted<T>` and `KeyComponent` are held back as reviewer decisions, the first over constant-time comparison of plaintext, the second because its `==` would publish a key-derivation collision as contract.

Require `Batch`'s resolver argument explicitly and declare its non-Sendability the way `Statement` does. Correct doc comments on `Statement.Options`, `CassandraSession.shutdown()` at both declaration sites, `Batch`, and the trace error-category constant.

### Result

Registering an encryption schema now states the columns the validation protects, and the four reasons a column fails to decode are distinguishable.

Source-breaking in two cases: callers that omitted `encryptedColumns` must pass a set, where `[]` preserves the previous behaviour; and any hand-written or retroactive `Hashable` conformance on one of the nine types is now redundant.

### DIFF
--- a/Sources/CassandraClient/Decoding.swift
+++ b/Sources/CassandraClient/Decoding.swift
@@ -81,6 +81,10 @@ extension CassandraClient {
         /// distinct causes into `nil` — a column the row lacks, a NULL value, a stored type that cannot
         /// convert, and a text column whose bytes are not valid UTF-8 — so probe the untyped accessor
         /// to tell them apart.
+        ///
+        /// The probe repeats the lookup the caller just made. That is deliberate: it runs only once that
+        /// lookup has already returned `nil`, so it never costs the success path, and classifying here
+        /// rather than at each call site keeps every one of them a single `throw`.
         private func columnFailure(_ type: Any.Type, forKey key: Key) -> Swift.DecodingError {
             guard let column: Column = self.row.column(key.stringValue) else {
                 return .columnMissing(forKey: key, codingPath: self.codingPath)
--- a/Sources/CassandraClient/Decoding.swift
+++ b/Sources/CassandraClient/Decoding.swift
@@ -39,11 +39,14 @@ extension CassandraClient {
         }
 
         public func unkeyedContainer() throws -> UnkeyedDecodingContainer {
-            throw DecodingError.notSupported("unkeyedContainer not supported")
+            throw DecodingError.unsupportedOperation("unkeyedContainer not supported", codingPath: self.codingPath)
         }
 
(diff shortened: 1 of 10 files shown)


Sources/CassandraClient/Decoding.swift
```diff
@@ -69,108 +72,138 @@ extension CassandraClient {
...
         }
 
+        /// The error for a typed column read that produced `nil`. The typed accessors collapse four
+        /// distinct causes into `nil` — a column the row lacks, a NULL value, a stored type that cannot
+        /// convert, and a text column whose bytes are not valid UTF-8 — so probe the untyped accessor
+        /// to tell them apart.
+        private func columnFailure(_ type: Any.Type, forKey key: Key) -> Swift.DecodingError {
+            guard let column: Column = self.row.column(key.stringValue) else {
+                return .columnMissing(forKey: key, codingPath: self.codingPath)
+            }
```
yifan-c, line 91: nit: add a comment to explain why the second/duplicated `row.column` call is there for the failure case. I think it is to avoid duplicating the common handling in the 8 callsites; meanwhile, the duplicated call is negligible in the (rare) failure case.

yifan-c (approved):

PR #131, opened 2026-09-11
Allow statements to be marked idempotent to enable host failover

Samsv77:
Adds an `isIdempotent` option to `Statement.Options`, which the bundled driver requires before it will retry a statement on another host or execute it speculatively.

### Motivation:

When a C* node drains during a rolling restart, it returns `OVERLOADED` state, this is handled gracefully in the vendored DataStax driver by connecting to the next host, but the Swift client never marks statements idempotent, so it skips the retry policy entirely. The caller gets a hard `serverOverloaded` while the rest of the cluster is healthy, and can lead to service crashing.

### Modifications:

- Added `isIdempotent: Bool?` to `Statement.Options`, mirroring the existing `Batch.Configuration.isIdempotent`. Keeping it optional preserves the distinction between leaving the driver default alone and explicitly setting it, which are different calls into the driver.
- Applied it through `cass_statement_set_is_idempotent` in the step that already applies `consistency` and `requestTimeout`. Both the plain and the prepared-statement initializers run that step, so both paths pick it up.
- Added an initializer that takes the new option. The original `init(consistency:requestTimeout:)` is kept with its signature and default arguments unchanged, forwarding to the new one. `isIdempotent` is deliberately required rather than defaulted on the initializer that accepts it, because if both overloads defaulted every shared parameter then `Options()` would be ambiguous. Its presence or absence now selects the overload.
- Included the setting in `Options.description`.
- Added `Tests/CassandraClientTests/StatementIdempotencyTests.swift`.

### Result:

Callers can mark a statement safe to replay:

```swift
let options = CassandraClient.Statement.Options(isIdempotent: true)
let statement = try CassandraClient.Statement(query: "SELECT id FROM test;", options: options)
```

When they do, an `OVERLOADED` or `SERVER_ERROR` response makes the driver hand the request to the next host in its query plan instead of failing the call, and a configured `speculativeExecutionPolicy` starts taking effect. Write timeouts also reach the retry policy, though the default policy still only retries batch-log writes.

### Tests

The flag was verified as reaching the driver by temporarily exposing Statement::is_idempotent() through a local C shim and reading it back off the request object. isIdempotent: true produces true, unset and false produce false, and disabling the setter as a negative control flips the positive assertions to failing. That scaffolding isn't part of this PR, since it needs a getter the public C API doesn't provide. The committed tests therefore pin the option's contract rather than the driver round-trip, and the failover it unlocks still needs a multi-node cluster with a draining node to observe end to end.

### DIFF
--- a/Sources/CassandraClient/Statement.swift
+++ b/Sources/CassandraClient/Statement.swift
@@ -234,6 +234,12 @@ extension CassandraClient {
             if let requestTimeout = options.requestTimeout {
                 try checkResult { cass_statement_set_request_timeout(self.rawPointer, requestTimeout) }
             }
+
+            if let isIdempotent = options.isIdempotent {
+                try checkResult {
+                    cass_statement_set_is_idempotent(self.rawPointer, isIdempotent ? cass_true : cass_false)
+                }
+            }
         }
 
         /// Encrypt plaintext and bind the result as bytes at the given parameter index.
@@ -514,6 +520,24 @@ extension CassandraClient {
             /// Sets the statement's request timeout in milliseconds. `nil` inherits
             /// ``CassandraClient/Configuration/requestTimeoutMillis``.
             public var requestTimeout: UInt64?
+            /// Whether the statement is safe to run more than once.
+            ///
+            /// A statement left unset is treated as unsafe to replay, which holds the driver back
+            /// from two recovery behaviors:
+            ///
+            /// - When a coordinator reports that it is overloaded, shutting down, or hit an internal
+            ///   error, the driver hands the request to the next host in its query plan instead of
+            ///   failing the call. Without this, a single node draining during a rolling restart
+            ///   surfaces as an error even though the rest of the cluster can serve the request.
+            /// - Speculative execution, configured by
+            ///   ``CassandraClient/Configuration/speculativeExecutionPolicy``, has no effect at all
+            ///   until a statement is marked safe to replay.
+            ///
+            /// - Important: Only mark a statement idempotent when running it twice leaves the same
+            ///   result as running it once. Counter updates, lightweight transactions, and appends to
+            ///   collections are not idempotent, and replaying one corrupts the value it touches.
+            ///   Plain reads, and writes that set fixed values, are.
+            public var isIdempotent: Bool?
 
             /// Type-erased backing store for ``encryptionContextBuilder``.
             private var _encryptionContextBuilder: (any Sendable)?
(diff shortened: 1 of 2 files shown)


Sources/CassandraClient/Statement.swift
```diff
@@ -562,16 +586,31 @@ extension CassandraClient {
                 self._encryptionContextBuilder != nil || self._encryptionTable != nil
             }
 
-            public init(consistency: CassandraClient.Consistency? = nil, requestTimeout: UInt64? = nil) {
+            public init(
+                consistency: CassandraClient.Consistency? = nil,
+                requestTimeout: UInt64? = nil,
+                isIdempotent: Bool?
```
yifan-c: How about `isIdempotent: Bool? = nil` as the third defaulted parameter, and delete the initializer, `public init(consistency: CassandraClient.Consistency? = nil, requestTimeout: UInt64? = nil)`? 

It is effectively the same.

Sources/CassandraClient/Statement.swift
```diff
@@ -562,16 +586,31 @@ extension CassandraClient {
...
+                self.isIdempotent = isIdempotent
+            }
+
+            /// Creates options that leave ``isIdempotent`` unset, so the driver keeps treating the
+            /// statement as unsafe to replay.
+            ///
+            /// `isIdempotent` is required rather than defaulted on the initializer that takes it, so
+            /// that passing it selects that initializer and omitting it selects this one.
+            public init(consistency: CassandraClient.Consistency? = nil, requestTimeout: UInt64? = nil) {
+                self.init(consistency: consistency, requestTimeout: requestTimeout, isIdempotent: nil)
```
yifan-c: delete this initializer and the doc coment

Tests/CassandraClientTests/StatementIdempotencyTests.swift
```diff
@@ -0,0 +1,117 @@
...
+    func testBothInitializersRemainAvailable() {
+        typealias Options = CassandraClient.Statement.Options
+        typealias Consistency = CassandraClient.Consistency
+
+        let withoutIdempotency: (Consistency?, UInt64?) -> Options = Options.init
+        let withIdempotency: (Consistency?, UInt64?, Bool?) -> Options = Options.init
+
+        XCTAssertNil(withoutIdempotency(.quorum, 1000).isIdempotent)
+        XCTAssertEqual(withIdempotency(.quorum, 1000, true).isIdempotent, true)
+    }
```
yifan-c: This test can be deleted.

Sources/CassandraClient/Statement.swift
```diff
@@ -514,6 +520,24 @@ extension CassandraClient {
...
+            ///   surfaces as an error even though the rest of the cluster can serve the request.
+            /// - Speculative execution, configured by
+            ///   ``CassandraClient/Configuration/speculativeExecutionPolicy``, has no effect at all
+            ///   until a statement is marked safe to replay.
+            ///
+            /// - Important: Only mark a statement idempotent when running it twice leaves the same
+            ///   result as running it once. Counter updates, lightweight transactions, and appends to
+            ///   collections are not idempotent, and replaying one corrupts the value it touches.
+            ///   Plain reads, and writes that set fixed values, are.
+            public var isIdempotent: Bool?
```
shruti-p-s, line 544: `Batch.add(prepared:parameters:options:)` takes a `Statement.Options`, but I don't think the driver reads this flag for a batch, the gates check `request()->is_idempotent()` on the  `BatchRequest`, which only `cass_batch_set_is_idempotent` sets. So setting it there compiles and does nothing, while the doc above reads as though it applies. Should we add a line pointing  batch callers at `Batch.Configuration.isIdempotent`?

Tests/CassandraClientTests/StatementIdempotencyTests.swift
```diff
@@ -0,0 +1,99 @@
...
+        )
+        XCTAssertEqual(options.isIdempotent, true)
+        XCTAssertEqual(options.consistency, .quorum)
+        XCTAssertEqual(options.requestTimeout, 1000)
+    }
+
+    /// The option is applied while the statement is being built, alongside consistency and the
+    /// request timeout. Building one for each setting pins that the apply step accepts them all
+    /// rather than rejecting a value the driver is fine with.
+    func testStatementConstructionAcceptsEachSetting() throws {
```
shruti-p-s: I think these two pass with the apply step removed; `cass_statement_set_is_idempotent` always returns `CASS_OK` (`statement.cpp:121`), so `checkResult` can't throw. I tried deleting the `if let isIdempotent` block locally and all six still pass. Should we drop these two?

yifan-c (approved):

PR #132, opened 2026-09-16
Public API for review ahead of 1.0 (not for merge)

shruti-p-s:
The public API of the package, one file per area, so it can be read and commented on line by line ahead of 1.0.

| File | Area |
|---|---|
| `0001-client-and-session.md` | Client, session, configuration, authentication |
| `0002-statements-and-execution.md` | Statements, prepared statements, batches, consistency |
| `0003-results-and-data-model.md` | `Rows`, `Row`, `Column`, `PaginatedRows` |
| `0004-errors.md` | `Error`, `ConfigurationError` |
| `0005-encryption.md` | Client-side encryption |
| `0006-observability.md` | `CassandraMetrics` |

Each renders its area as nested Swift, so a type's members and nested types sit inside its braces. Every public symbol lands in exactly one file. The areas are independent — comment on whichever you have a view on.

**Not for merge.** This exists to collect API review feedback; anything it settles lands as its own change, and this PR is then closed.

A few things about the rendering, so absences aren't misread:

- A protocol's body holds only its requirements. The default implementations and extension members the module supplies follow in an `extension`, so what a conformer must implement is distinguishable from what it gets for free.
- Inherited protocol members are listed, attributed to the protocol that declared them, but without their standard-library documentation. The declarations matter — a `Rows: Sequence` conformance hands adopters every buffering algorithm — but Swift's prose about `allSatisfy` doesn't.
- The symbol graph emits no `==` and no synthesized `encode(to:)`, though it emits `!=` and `init(from:)`. Those are reconstructed and marked `[reconstructed]`, so no member lacks a line to comment on.

The files are generated from the Swift symbol graph rather than written by hand, so they can be regenerated between rounds — `api-review/README.md` has the command. The renderer sits alongside them.


api-review/0001-client-and-session.md
```diff
@@ -0,0 +1,1005 @@
...
+    /// When `shared`, the `EventLoopGroup` is provided externally and its lifecycle will be managed by the caller.
+    /// When `createNew`, the library will create a new `EventLoopGroup` and manage its lifecycle.
+    public enum EventLoopGroupProvider : Sendable {
+        case createNew
+        case shared(EventLoopGroup)
+    }
+}
+
+/// API for executing statements against Cassandra.
+@preconcurrency public protocol CassandraSession : Sendable {
```
FranzBusch, line 739: Why does this protocol exist? It looks like it models the public API of the `CassandraClient`. We normally don't put such protocols in our packages and expect user's to abstract their dependencies for testing. I might be missing the purpose behind it though.

api-review/0001-client-and-session.md
```diff
@@ -0,0 +1,1005 @@
...
+/// `CassandraClient` is a wrapper around the [Datastax Cassandra C++ Driver](https://github.com/datastax/cpp-driver)
+///  and can be used to run queries against a Cassandra database.
+public final class CassandraClient : CassandraSession, Sendable {
+    /// Create a new instance of `CassandraClient`.
+    ///
+    /// - Parameters:
+    ///   - eventLoopGroupProvider: The ``EventLoopGroupProvider`` to use, uses ``EventLoopGroupProvider/createNew`` strategy by default.
+    ///   - configuration: The  client's ``Configuration``.
+    ///   - logger: The client's default `Logger`.
+    public init(eventLoopGroupProvider: EventLoopGroupProvider = .createNew, configuration: Configuration, logger: Logger? = nil)
```
FranzBusch, line 23: I think this should move on from using the EventLoopGroupProvider and accept and EventLoopGroup but default to the singleton one.

I also think this is potentially the wrong place to take a logger. Logger's should be passed down the call chain to ensure metadata correctly propagates. If this logger is used to log background activity then that's fine but it should definitely default to `.current`. More importantly all the other methods should take the task local `Logger.current` for logging to ensure the metadata is correct.

api-review/0001-client-and-session.md
```diff
@@ -0,0 +1,1005 @@
...
+
+    @available(macOS 15.0, iOS 18.0, visionOS 2.0, *)
+    public var encryptionSchemas: [String : CassandraClient.EncryptionSchema] { get }
+    @available(macOS 15.0, iOS 18.0, visionOS 2.0, *)
+    public var encryptor: CassandraClient.Encryptor? { get }
+    public var eventLoopGroup: EventLoopGroup { get }
+    public var keyspace: String? { get }
+    public let logger: Logger
```
FranzBusch, line 31: All these properties should have at least some public documentation but I am wondering if we really need them. Which ones do we actually except users to call?

api-review/0001-client-and-session.md
```diff
@@ -0,0 +1,1005 @@
...
+
+    /// Execute a batch of statements.
+    ///
+    /// - Parameters:
+    ///   - configuration: Options to apply to the batch.
+    ///   - logger: If `nil`, the client's default `Logger` is used.
+    ///   - build: Closure that adds statements to the batch.
+    @available(macOS 12, watchOS 8, iOS 15, tvOS 15, *)
+    public func batch(configuration: Batch.Configuration = .init(), logger: Logger? = .none, _ build: (inout Batch) async throws -> Void) async throws
```
FranzBusch, line 40: All these logger parameters should default to `.current`

api-review/0001-client-and-session.md
```diff
@@ -0,0 +1,1005 @@
...
+    /// Execute a batch of statements.
+    ///
+    /// - Parameters:
+    ///   - configuration: Options to apply to the batch.
+    ///   - eventLoop: The `EventLoop` to use, or create a new one.
+    ///   - logger: If `nil`, the client's default `Logger` is used.
+    ///   - build: Closure that adds statements to the batch.
+    public func batch(configuration: Batch.Configuration = .init(), on eventLoop: EventLoop? = .none, logger: Logger? = .none, _ build: (inout Batch) throws -> Void) -> EventLoopFuture<Void>
```
FranzBusch, line 48: On a general note, we should drop any Public API methods that return `EventLoopFuture`'s before tagging a 1.0.0. Nowdays we want to hide NIO as much as possible and the only thing where it should show up is in the init to pass the ELG.

api-review/0001-client-and-session.md
```diff
@@ -0,0 +1,1005 @@
...
+    public func execute(statement: sending Statement, pageSize: Int32, on eventLoop: EventLoop?, logger: Logger? = .none) -> EventLoopFuture<PaginatedRows>
+    public func getMetrics() -> CassandraMetrics
+    /// Create a new ``CassandraSession`` that can be used to perform queries on the given or configured keyspace.
+    ///
+    /// - Parameters:
+    ///   - keyspace: If `nil`, the client's default keyspace is used.
+    ///   - logger: If `nil`, the client's default `Logger` is used.
+    ///
+    /// - Returns: The newly created session.
+    public func makeSession(keyspace: String?, logger: Logger? = .none) -> CassandraSession
```
FranzBusch, line 204: Is a session a single connection to a Cassandra database?

api-review/0001-client-and-session.md
```diff
@@ -0,0 +1,1005 @@
...
+    /// Run insert / update / delete or DDL command where no result is expected.
+    ///
+    /// If `eventLoop` is `nil`, a new one will get created through the `EventLoopGroup` provided during initialization.
+    public func run(_ command: String, parameters: sending [CassandraClient.Statement.Value] = [], options: CassandraClient.Statement.Options = .init(), on eventLoop: EventLoop? = .none, logger: Logger? = .none) -> EventLoopFuture<Void>
+    /// Shutdown the client.
+    ///
+    /// - Note: It is required to call this method before terminating the program. `CassandraClient` will assert it was cleanly shut down as part of its deinitializer.
+    public func shutdown() throws
+    @available(macOS 12, watchOS 8, iOS 15, tvOS 15, *)
+    public func shutdownAsync() async throws
```
FranzBusch, line 311: Instead of this set of methods I think we should have a `withClient` API similar to what we did in Kafka + a single async `run` method that implements graceful shutdown using swift-service-lifecycle. This is how almost all the existing libraries in the ecosystem work.

api-review/0001-client-and-session.md
```diff
@@ -0,0 +1,1005 @@
...
+    ///   - closure: The closure to invoke, passing in the newly created session.
+    @available(macOS 12, watchOS 8, iOS 15, tvOS 15, *)
+    public func withSession(keyspace: String?, logger: Logger? = .none, closure: (CassandraSession) async throws -> Void) async throws
+    /// Create a new ``CassandraSession`` for the given or configured keyspace then invoke the closure.
+    ///
+    /// - Parameters:
+    ///   - keyspace: If `nil`, the client's default keyspace is used.
+    ///   - logger: If `nil`, the client's default `Logger` is used.
+    ///   - handler: The closure to invoke, passing in the newly created session.
+    public func withSession(keyspace: String?, logger: Logger? = .none, handler: (CassandraSession) throws -> Void) rethrows
```
FranzBusch, line 326: This seems duplicated to the below methods with typed throws. Can we get rid of this one?

api-review/0001-client-and-session.md
```diff
@@ -0,0 +1,1005 @@
...
+    /// }
+    /// ```
+    public protocol Authenticator : Sendable {
+        /// Answer a server challenge. The challenge is always present (possibly empty). Return `nil` when
+        /// the client has nothing further to send.
+        func evaluateChallenge(_ challenge: [UInt8]) throws -> [UInt8]?
+        /// The initial response that begins the SASL handshake. `nil` sends an empty response.
+        func initialResponse() throws -> [UInt8]?
+        /// Called once the server reports success, with any final token it sent (always present, may be empty).
+        func onSuccess(_ token: [UInt8]) throws
```
FranzBusch, line 384: What are implementors supposed to do here?

api-review/0001-client-and-session.md
```diff
@@ -0,0 +1,1005 @@
...
+        ///
+        /// - Throws: If a value is out of range or is not one of the accepted values for its key, or a required key is missing. `contactPoints` is
+        ///   validated here too, but because it is re-read it can also fail later, via the callback passed to ``contactPointsProvider``.
+        ///
+        /// - Parameters:
+        ///   - configReader: The reader to read configuration from.
+        ///   - logger: Logger for configuration warnings, such as SSL properties set while SSL is disabled
+        @available(macOS 15.0, watchOS 11.0, iOS 18.0, visionOS 2.0, tvOS 18.0, *)
+        public init(configReader: ConfigReader, logger: Logger) throws
+        @preconcurrency public init(contactPointsProvider: @escaping @Sendable (@escaping @Sendable (Result<ContactPoints, Swift.Error>) -> Void) -> Void, port: Int32, protocolVersion: ProtocolVersion)
```
FranzBusch, line 468: We should get rid of any `@preconcurrency` annotation before the 1.0.0 release.

api-review/0001-client-and-session.md
```diff
@@ -0,0 +1,1005 @@
...
+
+        /// A custom SASL authenticator. When set, it takes precedence over ``username``/``password``.
+        /// The instance is shared across all connections and invoked concurrently; see
+        /// ``CassandraClient/Authenticator``.
+        public var authenticator: (any CassandraClient.Authenticator)?
```
FranzBusch, line 473: Can we get all of the configuration options documented please.

api-review/0001-client-and-session.md
```diff
@@ -0,0 +1,1005 @@
...
+        public var compact: Bool?
+        public var connectTimeoutMillis: UInt32?
+        public var connectionHeartbeatInterval: UInt32?
+        public var connectionIdleTimeout: UInt32?
```
FranzBusch, line 477: Any time based configuration option should use `Duration`

api-review/0001-client-and-session.md
```diff
@@ -0,0 +1,1005 @@
...
+        ///   - logger: Logger for configuration warnings, such as SSL properties set while SSL is disabled
+        @available(macOS 15.0, watchOS 11.0, iOS 18.0, visionOS 2.0, tvOS 18.0, *)
+        public init(configReader: ConfigReader, logger: Logger) throws
+        @preconcurrency public init(contactPointsProvider: @escaping @Sendable (@escaping @Sendable (Result<ContactPoints, Swift.Error>) -> Void) -> Void, port: Int32, protocolVersion: ProtocolVersion)
+
+        /// A custom SASL authenticator. When set, it takes precedence over ``username``/``password``.
+        /// The instance is shared across all connections and invoked concurrently; see
+        /// ``CassandraClient/Authenticator``.
+        public var authenticator: (any CassandraClient.Authenticator)?
+        public var compact: Bool?
```
FranzBusch, line 474: Why is this an optional? What is the difference between `false` and `nil`?

api-review/0001-client-and-session.md
```diff
@@ -0,0 +1,1005 @@
...
+        public var connectTimeoutMillis: UInt32?
+        public var connectionHeartbeatInterval: UInt32?
+        public var connectionIdleTimeout: UInt32?
+        /// Sets the cluster's consistency level. Default is `.localOne`.
+        public var consistency: CassandraClient.Consistency?
```
FranzBusch, line 479: Should `Consistency` be nested inside `Configuration` instead of inside the client?

api-review/0001-client-and-session.md
```diff
@@ -0,0 +1,1005 @@
...
+        /// Provides the initial `ContactPoints` of the Cassandra cluster.
+        /// This can be a subset since each Cassandra instance is capable of discovering its peers.
+        public var contactPointsProvider: @Sendable (@escaping @Sendable (Result<ContactPoints, Swift.Error>) -> Void) -> Void
+        public var coreConnectionsPerHost: UInt32?
+        public var description: String { get }
```
FranzBusch, line 484: Not sure if the configuration needs to be custom string convertible

api-review/0001-client-and-session.md
```diff
@@ -0,0 +1,1005 @@
...
+        /// Poller cadence in milliseconds. Default `10000` (10s). `nil` or `0` disables the poller
+        /// while leaving ``metricsEnabled`` on; a `0` interval would busy-loop the poller.
+        /// Requires macOS 12 / iOS 15 or newer; on older platforms the poller does not start.
+        public var metricsPollIntervalMillis: UInt32?
+        /// Optional session name attached as a `session` dimension on every emitted metric.
+        /// Set this to disambiguate metrics when more than one metrics-enabled session runs in a
+        /// process, otherwise their identically-named gauges overwrite each other. `nil` = no dimension.
+        public var metricsSessionName: String?
+        public var numIOThreads: UInt32?
+        public var password: String?
```
FranzBusch, line 509: Isn't this solved with the authenticator?

api-review/0002-statements-and-execution.md
```diff
@@ -0,0 +1,290 @@
...
+
+        // [inherited from Equatable] 1 member
+        static func != (lhs: Self, rhs: Self) -> Bool
+
+        // [reconstructed] 1 member guaranteed by a conformance above that the symbol graph does not emit
+        public static func == (lhs: Self, rhs: Self) -> Bool
+    }
+
+    /// Consistency levels
+    public enum Consistency : Hashable, RawRepresentable, Sendable {
```
FranzBusch, line 68: All enums in the API need to be evaluated if they should be an enum and if they should be if they need to be `@nonexhaustive`. They should only be an enum if we expect users to switch over them. Otherwise a struct with static var's is better suited.

api-review/0002-statements-and-execution.md
```diff
@@ -0,0 +1,290 @@
...
+
+        // [compiler-synthesized] 1 member
+        init?(rawValue: String)
+
+        // [reconstructed] 1 member guaranteed by a conformance above that the symbol graph does not emit
+        public static func == (lhs: Self, rhs: Self) -> Bool
+    }
+
+    /// A server-side prepared statement that can be efficiently executed multiple times with different parameters.
+    public final class PreparedStatement : Sendable {
```
FranzBusch, line 96: Why is this a class?

api-review/0002-statements-and-execution.md
```diff
@@ -0,0 +1,290 @@
...
+
+        // [reconstructed] 1 member guaranteed by a conformance above that the symbol graph does not emit
+        public static func == (lhs: Self, rhs: Self) -> Bool
+    }
+
+    /// A prepared statement to run in a Cassandra database.
+    ///
+    /// Not `Sendable`: a `Statement` must not be used concurrently, or reused/mutated until a prior
+    /// `execute` has completed (the driver reads it asynchronously while the request is in flight).
+    public final class Statement : CustomStringConvertible {
```
FranzBusch, line 132: Why is this a class?

FranzBusch (commented):
Thanks for opening this. Very helpful to review the API again before an upcoming 1.0.0. I only got through some parts of the client and session so far but I feel like I accumulated enough comments to post them already. Some of them are very generally applicable across all the APIs and worth applying before doing more review.

One more note here is that I think we need to re-considering raising the minimum platform version of the package since all the `@availability` annotations are really adding up.

